PR #4, opened 2019-11-12
Add climate feature fan

glmnet:
See https://github.com/esphome/esphome/pull/845


gen-protoc.sh
```diff
@@ -4,4 +4,4 @@
 protoc --python_out=aioesphomeapi -I aioesphomeapi aioesphomeapi/*.proto
 
 # https://github.com/protocolbuffers/protobuf/issues/1491
-sed -i '' 's/import api_options_pb2 as api__options__pb2/from . import api_options_pb2 as api__options__pb2/' aioesphomeapi/api_pb2.py
+sed -i 's/import api_options_pb2 as api__options__pb2/from . import api_options_pb2 as api__options__pb2/' aioesphomeapi/api_pb2.py
```
OttoWinter: Ugh, this is an issue with different sed implementations - on my macOS system the extra '' is needed. I guess I will just alias sed to gsed to fix it here.

aioesphomeapi/api.proto
```diff
@@ -684,6 +698,7 @@ message ClimateStateResponse {
   float target_temperature_high = 6;
   bool away = 7;
   ClimateAction action = 8;
+  ClimateFanMode fan = 9;
```
OttoWinter: Let's call this `fan_mode` for consistency with HA.

aioesphomeapi/api.proto
```diff
@@ -702,4 +717,6 @@ message ClimateCommandRequest {
   float target_temperature_high = 9;
   bool has_away = 10;
   bool away = 11;
+  bool has_fan = 12;
+  ClimateFanMode fan = 13;
```
OttoWinter:
```suggestion
  ClimateFanMode fan_mode = 13;
```

aioesphomeapi/model.py
```diff
@@ -218,8 +232,10 @@ class ClimateInfo(EntityInfo):
 
 @attr.s
 class ClimateState(EntityState):
-    mode = attr.ib(type=ClimateMode, converter=ClimateMode, default=ClimateMode.OFF)
-    action = attr.ib(type=ClimateAction, converter=ClimateAction, default=ClimateAction.OFF)
+    mode = attr.ib(type=ClimateMode, converter=ClimateMode,
+                   default=ClimateMode.OFF)
+    action = attr.ib(type=ClimateAction, converter=ClimateAction,
+                     default=ClimateAction.OFF)
```
OttoWinter: I believe you need to add the new attributes (`fan_mode`) here.

Same for `supported_fan_modes` in the ClimateInfo object.

PR #8, opened 2020-05-07
Add support for fan direction

blejdfist:
This adds support for fan direction

Relevant pull requests
esphome/esphome-docs#580
esphome/esphome#1051


glmnet:
Well.. regarding the protoc generation I don't know why Otto is still generating those files with old python 2 utf headers.

Anyway, this is stuck without him, I am not really expert in python, but even if we commit to aioesphome this, and bump version, the library will not reach pyp magically (I guess) and we will go no further than that. Am I right with this reasoning?

blejdfist:
@glmnet Maybe it was just an older version of the protobuf compiler? I used version 3.6.1 (current version in Ubuntu 20.04).

You are probably right, if this is merged it will not reach pypi automatically.
There doesn't seem to be any automation in this repo.

What is your opinion on merging esphome/esphome#1051 (+ documentation) in the meanwhile?
It does not require the changes to aioesphome to be merged at the same time since protobuf will simply ignore fields it does not know about).

blejdfist:
@glmnet It seems like the utf-8 headers in the generated files are now gone on master as well so the diff is cleaner now.

glmnet:
@blejdfist seems like your code finally went thru, now you'll need to re-open the PR at HA

blejdfist:
@glmnet That's great, I was starting to loose hope :)
I'll look into it as soon as I can.

PR #13, opened 2020-08-11
Add the ability to use the shared Home Assistant Zeroconf instance

bdraco:


bdraco:
@glmnet or @OttoWinter If you have a moment, could we get a release bump with this fix.  I'm trying to close out https://github.com/home-assistant/core/issues/38862 and https://github.com/home-assistant/core/pull/38747 and this is the one of the last ones to be resolved before we start warning about duplicate Zeroconf instances in Home Assistant.

PR #21, opened 2020-11-29
Added Initial Support for Climate Presets

fraschizzato:
Added Initial support for Climate Presets. Original version had only "away", now it supports also "boost" and "sleep".
Commented support for "none" is provided but it's maintained the initial logic of "home" for normal operation (no presets).
Tested with Bang-Bang component and seems to work fine both with official esphome Home Assistant integration and with that modified esphome component: https://github.com/fraschizzato/ha-esphome-climate/tree/master
At this stage only with that custom component is possible to use the "boost" and "sleep" function. With the previous (official) one it works everything as before


glmnet:
Can you run the codegen again?

fraschizzato:
> Can you run the codegen again?

Sorry, at this time I'm out-of-sync from that project. Can you be more specific on what I have to do? 
I've tried to re-run gen-protoc.sh but with the same output because my local repo was not modified.
I've now seen a conflict with api_pb2.py but didn't know how to fix this. Can you give me some hint?
Thanks

fraschizzato:
Ok, maybe now i'm "on-sync" XD, tell me if I'm wrong.
Note that this pull is useless without: https://github.com/home-assistant/core/pull/43739 and https://github.com/esphome/esphome/pull/1401
However is not a breaking change

aioesphomeapi/api.proto
```diff
@@ -710,6 +712,9 @@ message ListEntitiesClimateResponse {
   bool supports_action = 12;
   repeated ClimateFanMode supported_fan_modes = 13;
   repeated ClimateSwingMode supported_swing_modes = 14;
+  bool supports_boost = 15;
+  bool supports_night = 16;
```
Lumpusz, line 716: This is a bit inconsistent. It is for "SLEEP" preset, but here it is called "night". Night mode is when the power for the compressor in the outdoor unit is reduced to make it more silent so it complies with local regulations during nighttime. Sleep is when the fan is more silent in the indoor unit, so people in the room can sleep more comfortably.

fraschizzato: Didn't remember in wich stage (aioesphome,ephome or home-assistant) I got problems, but using the word "SLEEP" that was assumed as "sleep function" in the code. To workaround this, I've used the word "night" in all three integrations to maintain coerency. My device reports the same for Sleep/Night. By the way, at this time, home-assistant handles (in main climate component) the "SLEEP" mode as "Fan Silent Mode", so it can't be confused. 
Also the "boost" mode is renamed, in that case accordingly to home-assistant, the correct match for many devices had to be "Turbo". 
At this time I will not change all for a simple variable/function name that has nothing to do with functionality, with the risk of breaking a fully working mod and/or a waste of time. Expecially in a project that every day seems more abandoned. When an outdoor control will be added there will be 2 ways: renaming that night mode or simply add an "outdoor" note in variable/function name, didn't see the problem. Feel free to fork, change and request merge, but remember: Maintaining existing naming structure, on all three integrations, causes a bug when the word "sleep" is used, somewhere is interpreted as instruction instead of variable.

glmnet: is there somewher a translation of `SLEEP` -> `NIGHT`? that will be very confusing, if `SLEEP` is a keyword somewhere it should be avoided by replacing by e.g. `SLEEP_`

glmnet: Where is the con


> I got problems, but using the word "SLEEP" that was assumed as "sleep function" in the code.

Most likely in the code generator, notice however we do have a component named `switch` which is already renamed to `switch_` to avoid this same problem.

fraschizzato: The preset is always referenced as "night" in all 3 codes, only in HomeAssistant component it's linked to the "PRESET_SLEEP" variable, however that variable in HomeAssistant (at least in my language) is then handled as "Notte" == "Night". In the same code the "PRESET_NONE" is mapped as "PRESET_HOME" to maintain the previous esphome logic (derived from BangBang climate), to avoid Breaking Change. 

Note: At this time, after latest (previous requested) codegen the component doesn't work for me (have to revert in local repo). Didn't know if it's related only to version numbering. In HomeAssistant i get an error about "supported_night" (but is the first in declaration) not found in aioesphomeapi ClimateInfo, but I've double checked that is there. In next week I will try to check if only a version numbering issue (that wrongly updates via pip).

fraschizzato: Now I will try to check if all is working after changing night with sleep

fraschizzato:
The real error was related to HA dependencies. Now both custom_components and HA PR require aioesphomeapi 2.6.5 (that still not exists because on PR). So now HA PR is broken due to lack of aioesphomeapi. 

Then: fixing night/sleep stupid thing I have to deal with the fact that lint does not allow to use the "_" everywhere so it will be a mix of "sleep","sleep_", and "sleep__" (because to maintain code coerency some functions are already declared with "_"). Plus, before the "sleep" thing, if anyone have to deal with the "Climate sleep function" had to simply search for "night", now, if anyone searches "sleep" it will be prompted to climate, delay, switches, etc.... Good luck to everyone

aioesphomeapi/api.proto
fraschizzato: all seems working fine
set version at 2.6.5 to avoid cross-dependency in ha component

PR #24, opened 2021-01-06
Climate

Lumpusz:
Update esphome climate schema to be in sync with HA climate schema, intended to be backward compatible, so migration can happen incrementally. This PR is part of a bundle, as aioesphomeapi, HA esphome component also needs to change. Other relate PRs:
- https://github.com/esphome/esphome/pull/1471
- https://github.com/home-assistant/core/pull/48116


fraschizzato:
It has been tested? Can't see in commits/code how is that handled by esphome HA components (homeassistant/core). Expecially how it handles the deprecation warning and the actual "if not-away = home" statement in HA side (breaking change for many esphome components - ie BangBangClimate) . Also in esphome/esphome code I can't see anything related (or available for tests), can see something in midea-preset branch but not implemented coerently to aioesphomeapi.
Can't even evaluate if working/fittable/implementable. 
Knowing that aioesphomeapi is the first step of the three needed, it's also needed to see (from now) how all works togheter. I don't think is a good idea to merge a piece that - maybe - needs some adjustement to fit the other 2.

Lumpusz:
> It has been tested? Can't see in commits/code how is that handled by esphome HA components (homeassistant/core). Expecially how it handles the deprecation warning and the actual "if not-away = home" statement in HA side (breaking change for many esphome components - ie BangBangClimate) . Also in esphome/esphome code I can't see anything related (or available for tests), can see something in midea-preset branch but not implemented coerently to aioesphomeapi.
> Can't even evaluate if working/fittable/implementable.
> Knowing that aioesphomeapi is the first step of the three needed, it's also needed to see (from now) how all works togheter. I don't think is a good idea to merge a piece that - maybe - needs some adjustement to fit the other 2.

Thanks for your input. This PR is there for the main devs to review the approach. This is indeed the first step, and if they have fundamental concerns, I would like to address them before the rest of the components are updated. Stay tuned for updates.

aioesphomeapi/model.py
```diff
@@ -245,6 +255,13 @@ def _convert_climate_swing_modes(value):
     return [ClimateSwingMode(val) for val in value]
 
 
+def _convert_climate_presets(value):
+    return [ClimatePreset(val) for val in value]
+
+
+def _convert_string(value):
+    return [val for val in value]
```
Lumpusz, line 263: @glmnet lint complains about this line, however without this, HA fails to persist the object as JSON, as this object returned from the protobuf protocol is not a list, but some kind of google iterable, so here we just convert the iterable to a standard python list, so JSON serializer can work with it. How we should solve this lint check?

glmnet: https://stackoverflow.com/questions/28829236/is-it-possible-to-ignore-one-single-specific-line-with-pylint/29562085

setup.py
```diff
@@ -10,7 +10,7 @@
     long_description = readme_file.read()
 
 
-VERSION = '2.6.4'
+VERSION = '2.6.5'
```
glmnet, line 13: version change will be on another commit

aioesphomeapi/model.py
```diff
@@ -245,6 +255,15 @@ def _convert_climate_swing_modes(value):
...
 
 
+def _convert_climate_presets(value):
+    return [ClimatePreset(val) for val in value]
+
+
+def _convert_string(value):
+    # need to convert protobuf iterable to list, otherwise HA fails to serialize it as JSON
+    # pylint: disable=R1721
+    return [val for val in value]
```
bdraco, line 265:
```suggestion
    return list(value)
```

jesserockz:
@Lumpusz Can you update this PR?

Lumpusz:
> @Lumpusz Can you update this PR?

Done.

aioesphomeapi/model.py
jesserockz: @Lumpusz Have you tried this suggestion?

Lumpusz: I have not tried this, did not think it was important. This is due to the fact that lint thinks it is already a list, but it is some kind of proto specific class, and without this HA fails.

Lumpusz:
@jesserockz When can we merge this PR? How do you plan to orchestrate the release, as the PR for HA component is still pending, this one has to go in first, increment version, and update the HA PR.

jesserockz:
Yeah I am aware. I have it on my list to look at and test and just got busy with other stuff for the esphome beta. After I merge I will bump the version and release and then you can update the HA pr. 

Thanks

aioesphomeapi/model.py
```diff
@@ -270,6 +289,15 @@ class ClimateInfo(EntityInfo):
     supported_swing_modes = attr.ib(
         type=List[ClimateSwingMode], converter=_convert_climate_swing_modes, factory=list
     )
+    supported_custom_fan_modes = attr.ib(
+        type=List[str], converter=_convert_string, factory=list
```
OttoWinter, line 293:
```suggestion
        type=List[str], converter=list, factory=list
```

Just as an FYI - the `_convert_string` is equivalent to `list` converter.

aioesphomeapi/model.py
```diff
@@ -289,6 +317,15 @@ class ClimateState(EntityState):
     swing_mode = attr.ib(
         type=ClimateSwingMode, converter=ClimateSwingMode, default=ClimateSwingMode.OFF
     )
+    custom_fan_mode = attr.ib(
+        type=str, default=None
+    )
+    preset = attr.ib(
+        type=ClimatePreset, converter=ClimatePreset, default=ClimatePreset.HOME
```
OttoWinter, line 324:
```suggestion
        type=ClimatePreset, converter=ClimatePreset, default=ClimatePreset.ECO
```

See comment at the top of the file:
```py
# All fields in here should have defaults set
# Home Assistant depends on these fields being constructible
# with args from a previous version of Home Assistant.
# The default value should *always* be the Protobuf default value
# for a field (False, 0, empty string, enum with value 0, ...)
```

aioesphomeapi/model.py
```diff
@@ -289,6 +317,15 @@ class ClimateState(EntityState):
     swing_mode = attr.ib(
         type=ClimateSwingMode, converter=ClimateSwingMode, default=ClimateSwingMode.OFF
     )
+    custom_fan_mode = attr.ib(
+        type=str, default=None
```
OttoWinter, line 321:
```suggestion
        type=str, default=''
```

aioesphomeapi/model.py
Lumpusz: Shouldn't in that case HOME be the default value? HOME essentially means normal operation, a.k.a. no special preset. If we put ECO as default, what would happen with devices that don't even have ECO?

OttoWinter: That depends on the HA implementation.

The main problem is: if `preset` is not set in `api_connection.cpp`, or with older esphome client versions. `aioesphomeapi` will _receive_ that as `0` aka `ECO` mode (lack of a field in protobuf means use its 0 value). `default` is not really used here and only for HA to be able to restore entity infos.

ECO as 0 value _would_ be ok if the HA implementation checks if the preset does actually exist in `self.preset_modes` ([here](https://github.com/home-assistant/core/pull/48116/files#diff-d335f2f1f824951421ef024c3be5ef2aee032239bc4bbb723b8a5b54814ae889R281).

But that's not really a clean solution (and any other api clients would have to do the same). So yes it would be best if we encode `HOME` as 0 (would be an API protocol change, but I think it's ok if we change it soon)

OttoWinter: You can replicate it like this: use an old esphome version (1.18.0), or a climate device with neither `preset` nor `away` traits. Then HA will show the climate preset as `ECO`

jesserockz: I too think if HOME should be default then move it to 0 on both sides. Since its not actually used yet we can safely change it on ESPHome side

Lumpusz: Ok, let's do that, I will have some time to make the changes next week.

Lumpusz: @jesserockz  @OttoWinter Changes to make HOME 0 are ready, please review

jesserockz:
Replaced by #42. Thanks for keeping this going @Lumpusz

PR #31, opened 2021-05-23
Add state class

frenck:
This add support for Home Assistant's Sensor attribute:

- `state_class`

This is a first step in adding support for it, by extending the protocol.


aioesphomeapi/api.proto
```diff
@@ -424,6 +424,8 @@ message ListEntitiesSensorResponse {
   int32 accuracy_decimals = 7;
   bool force_update = 8;
   string device_class = 9;
+  string state_class = 10;
+  string last_reset = 11;
```
OttoWinter, line 428: What is last_reset?

* If it's a dynamic attribute (changes across time), it should not be in this proto message. The List* messages are properties that describe the entity's static data, and for example in HA are persisted to disk.

* If this a datetime, it should be sent as an integer.

* It looks like this attribute is just there for some integrations like utility meter (correct me if i'm wrong). I don't think we should change ESPHome's sensor model just for making it work with one or two very Home Assistant specific integrations. For ESPHome, sensors only do one thing, and that is to represent a floating point value, and I think that simplicity is good. Keep in mind adding atteibutes to a core component like Sensor would also i crease the memory usage for every single integration (no matter if they use it or not).

frenck: > If it's a dynamic attribute (changes across time), it should not be in this proto message. 

It can reset, for HA there is no difference. Please suggest an alterantive.

> If this a datetime, it should be sent as an integer.

It is an optional datetime, setting it to 0 is actually a valid one (and thus triggers logic), so hence I took this approach. An valid alternative is welcome.

> It looks like this attribute is just there for some integrations like utility meter (correct me if i'm wrong).

It is being implemented at this moment, it is for gathering long-term stats in the recorder (downsampling data).

> For ESPHome, sensors only do one thing, and that is to represent a floating point value, and I think that simplicity is good.

It signals Home Assistant to gather stats over sensors in the long term. Considering ESPHome provides a lot of sensors (just look at temperature), this really beneficial, otherwise, all ESPHome sensors would just be a current value displayed for the future.

frenck: > It looks like this attribute is just there for some integrations like utility meter

That integration has been adjusted as well.

frenck: Moved `last_reset` to `SensorState`

frenck:
`state_class` can signal a `measurement` to Home Assistant, which triggers stats gathering over time, including downsampling.

For some sensors, they don't represent "current" state in time, but instead, count over X time. The usage of those sensors still can be useful and extracted.

So, `last_reset` has value for things like counters & meters. It will tell Home Assistant when the state value was the last reset. Based on that, it will try to extract things like usage data for storing stats over a longer period of time.

At this moment in ESPHome, Pulse Meter, Pulse Counter, Total Daily Energy and Template Sensor are good candidates to set (or to allow setting) this property.

aioesphomeapi/api.proto
OttoWinter: > It is being implemented at this moment, it is for gathering long-term stats in the recorder (downsampling data).

I see; can you point me to where this is being implemented? I'm still not clear on both what this value is supposed to mean (what is "reset"? when the device/sensor last restarted?) and how that would impact downsampling (sensor values are just datapoints at a singular point in time and phyiscal sensors generally don't care when they were last reset, as they only give you the value of the current moment; I don't understand how/why a "reset" would matter for downsampling).

> It is an optional datetime, setting it to 0 is actually a valid one (and thus triggers logic), so hence I took this approach. An valid alternative is welcome.

I think 0 would be perfectly fine. In fact we could consider any datetime <2021-05-23 as illegal, because (as far as I understand the feature) such values can not happen without a bug. Alternatively, adding a second boolean that indicates the presence (like with light commands for example) would be ok, but only if there's a legitimate case where a dt of `0` would be valid.

> Moved last_reset to SensorState

Ok, that's better - depending on what the semantics of this property are, it could also be a device-wide property (all sensors are "reset" when the device restarts, and I can't think of a case where a sensor restarts without the device restarting).

frenck: last_reset = 0 is epoch start, which is considered an non-resettable total. Thus not a correct value to send.

OttoWinter:
> For some sensors, they don't represent "current" state in time, but instead, count over X time. The usage of those sensors still can be useful and extracted.

I see, so `last_reset` is useful for sensors that represent the integral of a sensor value over time. Thanks for explaining!

aioesphomeapi/api.proto
frenck: > I see; can you point me to where this is being implemented? 

Quite a few PRs at the Home Assistant Core. I guess the best is to look at the PRs done by emontnemery last weeks on that repo.

Some key PRS:
- <https://github.com/home-assistant/core/pull/49852>
- <https://github.com/home-assistant/core/pull/50829>

frenck: Converted value to an integer, setting the default to `-1`.

frenck:
I've removed `last_reset` for now, implementing `state_class` is a good first change. Gives a bit of time to talk/discuss `last_reset` first.

PR #36, opened 2021-06-17
Switch from attrs to dataclasses

OttoWinter:
Python [attrs](https://www.attrs.org/en/stable/) was used in the time when we still had to work with python 3.5

In Python 3.7+, we have python's stdlib [dataclasses](https://docs.python.org/3/library/dataclasses.html).

They're mostly the same (but `dataclass` lacks `converter` functionality, which is manually added with a metaclass), but using the stdlib feels cleaner.

Note: this is a pretty big API breaking change, so would need major version bump, and some changes in HA.


aioesphomeapi/client.py
```diff
@@ -284,7 +256,7 @@ def on_msg(msg: message.Message) -> None:
             if isinstance(msg, CameraImageResponse):
                 data = image_stream.pop(msg.key, bytes()) + msg.data
                 if msg.done:
-                    on_state(CameraState(key=msg.key, image=data))
```
jesserockz, line 287: The issue people are experiencing on HA 2021.7 is because the new `from_pb` function only matches the attributes in the dataclass by exact name. That is `image` in `CameraState` but in `CameraImageResponse` it is `data`. Simple fix is to rename the attr to data and the references inside HA

PR #42, opened 2021-06-21
Climate preset, custom fan mode and custom preset

OttoWinter:
Updates #24 for changes in https://github.com/esphome/esphome/pull/1941

(for some reason I could not push directly to #24, so opening as a new PR)

Also implements https://github.com/esphome/esphome/pull/1951

# Breaking Change

* ClimateMode `AUTO` has been renamed to `HEAT_COOL` (and a new `AUTO` mode was added), this is for consistency between HA and esphome
* `away` for climate is removed. Instead, it is now encoded in `preset`. For compatibility with older peers use `ClimateInfo.supported_presets_compat` and `ClimateState.preset_compat`


aioesphomeapi/api.proto
```diff
@@ -704,6 +705,16 @@ enum ClimateAction {
   CLIMATE_ACTION_DRYING = 5;
   CLIMATE_ACTION_FAN = 6;
 }
+enum ClimatePreset {
+  CLIMATE_PRESET_NONE = 0;
```
Lumpusz, line 709: Shoudn't this be removed, and HOME is to be set to 0?

aioesphomeapi/model.py
```diff
@@ -286,6 +287,17 @@ class ClimateAction(APIIntEnum):
     FAN = 6
 
 
+class ClimatePreset(APIIntEnum):
+    NONE = 0
```
Lumpusz, line 291: This one too.

aioesphomeapi/api.proto
OttoWinter: Correct, but this PR also implements https://github.com/esphome/esphome/pull/1951 (adds `NONE` preset, because HA has it too and NONE would be a good 0 value for an enum; for example if you have a device with say only eco and boost presets, but not home/away)

OttoWinter: Huh, you're right - weird that it's missing there.

Anyway, adding it can't hurt - and now's the best time to add it without a breaking change

Lumpusz: Indeed, I can see that this was introduced in this commit: https://github.com/home-assistant/core/commit/50d4921d0a2ddee4503e9e260cd65a767902cadd#diff-bc3dca6e297a377d1bb686440fe7891a93ec7f87c3a5fcaa8dca8aa3bf157fac, but they most likely forgot to update the docs.

Lumpusz: Also, I see that in https://github.com/esphome/esphome/pull/1941 HOME was still the default with 0, so that may need to be updated as well.

PR #47, opened 2021-06-28
Add number entities

jesserockz:
- esphome/esphome#1971
- home-assistant/core#52241


OttoWinter (commented):
This side is ok I think.

Minor thing that might be worth exploring: Are there use-cases where `min`/`max`/`step` are dynamic? If so we can move those attributes to `NumberState`

jesserockz:
I cant think of any cases where I would want them to be dynamic. I would either define them in yaml (for the template), or other platform code would.

OttoWinter (approved):

PR #63, opened 2021-07-09
Fix camera not merging packets

OttoWinter:
Bug originally introduced here: https://github.com/esphome/aioesphomeapi/pull/36/files#diff-753f76689d07a39d2daa897bd6416dcb0546b60d15c3a944768fa4b8a75fd869L287

This fix fixed the protobuf->dataclass conversion: https://github.com/esphome/aioesphomeapi/pull/62/files but was incomplete.

For `on_state` we need to return the _merged_ data across the different image responses received.

Related: https://github.com/home-assistant/core/issues/52770


jesserockz:
:man_facepalming:

PR #64, opened 2021-07-10
Add pytest unit testing

OttoWinter:
After #62 and #63 I think it's good to have some unit testing in place.

This PR adds unit testing for most of the codebase except `connection.py` (code is hard to test) and `reconnect_logic.py`

Tests are currently only run on python 3.8+ (not 3.7) because they need [`unittest.mock.AsyncMock`](https://docs.python.org/3.8/library/unittest.mock.html#unittest.mock.AsyncMock) - future work might include backporting `AsyncMock` to also get test coverage with 3.7+

### DIFF
--- a/aioesphomeapi/model.py
+++ b/aioesphomeapi/model.py
@@ -168,9 +168,9 @@ class CoverState(EntityState):
     )
 
     def is_closed(self, api_version: APIVersion) -> bool:
-        if api_version >= APIVersion(1, 1):
-            return self.position == 0.0
-        return self.legacy_state == LegacyCoverState.CLOSED
+        if api_version < APIVersion(1, 1):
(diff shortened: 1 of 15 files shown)


aioesphomeapi/model.py
```diff
@@ -398,7 +398,7 @@ class ClimateState(EntityState):
     )
     custom_fan_mode: str = ""
     preset: Optional[ClimatePreset] = converter_field(
-        default=ClimatePreset.HOME, converter=ClimatePreset.convert
+        default=ClimatePreset.NONE, converter=ClimatePreset.convert
```
OttoWinter, line 401: This was a small bug caught by the tests, the default should be 0 `=NONE`

PR #74, opened 2021-07-23
Color mode implementation

oxan:
API changes for esphome/esphome#2012.


aioesphomeapi/model.py
```diff
@@ -207,26 +207,46 @@ class FanState(EntityState):
...
-    supports_white_value: bool = False
-    supports_color_temperature: bool = False
+    supported_color_modes: List[LightColorMode] = converter_field(
+        default_factory=list, converter=list
+    )
     min_mireds: float = 0.0
     max_mireds: float = 0.0
     effects: List[str] = converter_field(default_factory=list, converter=list)
 
+    # deprecated, do not use
```
OttoWinter, line 231: Let's call these `legacy_...`.

Then let's add a `supported_color_modes_compat` method that assembles the `color_mode` values given the peer API version (like with climate presets). Same for the call in `client.py`.

Advantage then is that we have all the compat logic in aioesphomeapi, and not in HA.

aioesphomeapi/model.py
```diff
@@ -207,26 +207,46 @@ class FanState(EntityState):
...
+
+
 @dataclass(frozen=True)
 class LightInfo(EntityInfo):
     supports_brightness: bool = False
-    supports_rgb: bool = False
-    supports_white_value: bool = False
-    supports_color_temperature: bool = False
+    supported_color_modes: List[LightColorMode] = converter_field(
+        default_factory=list, converter=list
```
OttoWinter, line 225:
```suggestion
        default_factory=list, converter=LightColorMode.convert_list
```

To convert the values from the protobuf enums to our python enums

aioesphomeapi/model.py
OttoWinter: Ok scratch that about `client.py` - the client class doesn't have enough info to transform those calls. But for `LightInfo` we can do it:

```python3
    legacy_supports_rgb: bool = False
    legacy_supports_white_value: bool = False
    legacy_supports_color_temperature: bool = False

    def supported_color_modes_compat(self, api_version: APIVersion) -> List[LightColorMode]:
        if api_version < APIVersion(1, 6):
            key = (
                self.legacy_supports_rgb, 
                self.legacy_supports_white_value, 
                self.legacy_supports_color_temperature
            )
            # map legacy flags to color modes, 
            # key: (rgb, white, color_temp)
            modes_map = {
                (False, False, False): [],
                (False, False, True): [LightColorMode.COLOR_TEMPERATURE],
                (False, True, False): [],  # invalid
                (False, True, True): [],  # invalid
                (True, False, False): [LightColorMode.RGB],
                (True, False, True): [],  # invalid
                (True, True, False): [LightColorMode.RGB_WHITE],
                (True, True, True): [LightColorMode.RGB_COLOR_TEMPERATURE]
            }
            return modes_map[key]

        return self.supported_color_modes
```

oxan: > Let's call these `legacy_...`.

Just to be sure, we don't care about backward compatibility here?

OttoWinter: Not really, the package is only really used by HA afaik (and some smaller things, but none that would touch color modes I think). We can just create a major version bump.

OttoWinter (approved):

aioesphomeapi/model.py
```diff
@@ -207,26 +207,73 @@ class FanState(EntityState):
...
+            return modes_map[key] if key in modes_map else []
+
+        return self.supported_color_modes
+
 
 @dataclass(frozen=True)
 class LightState(EntityState):
     state: bool = False
     brightness: float = 0.0
+    color_mode: LightColorMode = LightColorMode.UNKNOWN
```
OttoWinter, line 268:
```suggestion
    color_mode: Optional[LightColorMode] = converter_field(
        default=LightColorMode.UNKNOWN, converter=LightColorMode.convert
    )
```

PR #86, opened 2021-08-06
Fix old lights with color temperature but no white support

oxan:
ESPHome never supported this, and no built-in lights used this, but apparently it was used by third-party components. Without this fix, those lights would show up as on-off lights in HA. Make them RGBCT lights, which doesn't fit perfectly as they don't expect a white value, but they already had that in pre-2021.8 HA anyway and it's better than being totally broken.


OttoWinter (approved):
Thanks!

PR #95, opened 2021-08-23
Switch light colormode to be bitfields instead of enums

OttoWinter:
See also https://github.com/esphome/esphome/pull/2186

This should make it possible to add new capabilities without breaking clients.

When a client is looking for suitable a color mode, it should go through all color modes listed in `supported_color_modes` and select the best one that supports all features requested in the call.

Protobuf encoding is the same (no breaking change there), enums are varints so like int32 (or int64)


jesserockz (approved):

jesserockz:
@OttoWinter 

I took a look at the HA side for this today and realised we still want/need a ColorMode type class.

I ended up adding this in `light.py`, but wondering if it should be in here in the library?

The reason being that when we want to control the light, we need to set the `color_mode` that the light should be/change to. And without this, we end up with logical `or` everywhere to build the color_mode int.

```python
class ESPHomeColorMode:
    UNKNOWN = 0
    ON_OFF = LightColorCapability.ON_OFF
    MONOCHROMATIC = ON_OFF | LightColorCapability.BRIGHTNESS
    WHITE = MONOCHROMATIC | LightColorCapability.WHITE
    COLOR_TEMPERATURE = MONOCHROMATIC | LightColorCapability.COLOR_TEMPERATURE
    COLD_WARM_WHITE = MONOCHROMATIC | LightColorCapability.COLD_WARM_WHITE
    RGB = MONOCHROMATIC | LightColorCapability.RGB
    RGBW = RGB | WHITE
    RGBWW = RGB | COLD_WARM_WHITE
    RGB_COLOR_TEMPERATURE = RGB | COLOR_TEMPERATURE | WHITE
```

Issue #87, opened 2021-08-06
Can't connect after latest update "A Stream Socket was expected" Error


v-alvex:
`Traceback (most recent call last):                                                                                        
File "C:\Users\Administrator\aio\aio\lib\site-packages\aioesphomeapi\client.py", line 176, in connect                     
await self._connection.connect()                                                                                      
File "C:\Users\Administrator\aio\aio\lib\site-packages\aioesphomeapi\connection.py", line 157, in connect                 self._socket_reader, self._socket_writer = await asyncio.open_connection(                                             
File "C:\Users\Administrator\AppData\Local\Programs\Python\Python38\lib\asyncio\streams.py", line 52, in open_connection                                                                                                                          transport, _ = await loop.create_connection(                                                                          
File "C:\Users\Administrator\AppData\Local\Programs\Python\Python38\lib\asyncio\base_events.py", line 1047, in create_connection                                                                                                                  
raise ValueError(                                                                                                   
ValueError: A Stream Socket was expected, got <socket.socket fd=592, family=AddressFamily.AF_INET, type=0, proto=6, laddr=('192.168.2.4', 32149), raddr=('192.168.2.25', 6053)>                                
The above exception was the direct cause of the following exception:
Traceback (most recent call last):                                                                                        
File "t.py", line 24, in <module>                                                                                         
loop.run_until_complete(main())                                                                                       
File "C:\Users\Administrator\AppData\Local\Programs\Python\Python38\lib\asyncio\base_events.py", line 616, in run_until_complete                                                                                                                  
return future.result()                                                                                                
File "t.py", line 10, in main                                                                                             
await api.connect(login=True)                                                                                         
File "C:\Users\Administrator\aio\aio\lib\site-packages\aioesphomeapi\client.py", line 184, in connect                     
raise APIConnectionError(                                                                                           
aioesphomeapi.core.APIConnectionError: Unexpected error while connecting to 192.168.2.25: 
A Stream Socket was expected, got <socket.socket fd=592, family=AddressFamily.AF_INET, type=0, proto=6, laddr=('192.168.2.4', 32149), raddr=('192.168.2.25', 6053)>`

After replacing, in line 132 of connection.py

`self._socket = socket.socket( family=addr.family, type=addr.type, proto=addr.proto )`

with

`self._socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)`

The problem is fixed.

OttoWinter:
[...]

PR #105, opened 2021-09-17
fixes connection isssue#87

martgras:
Fixes the connection error under Windows : https://github.com/esphome/aioesphomeapi/issues/87
SOCK_STREAM is required anyways so make sure we get only SOCK_STREAM back from getaddrinfo


OttoWinter (approved):
Thanks!

PR #122, opened 2021-10-16
Expect a name for connections

OttoWinter:
Part of the fix for https://github.com/home-assistant/core/issues/52239

~~Not fully tested yet~~


aioesphomeapi/_frame_helper.py
```diff
@@ -178,18 +179,35 @@ async def _read_frame(self) -> bytes:
...
         chosen_proto = server_hello[0]
         if chosen_proto != 0x01:
             raise HandshakeAPIError(
                 f"Unknown protocol selected by client {chosen_proto}"
             )
 
+        # Check name matches expected name (for noise sessions, this is done
+        # during hello phase before a connection is set up)
+        # Server name is encoded as a string followed by a zero byte after the chosen proto byte
+        server_name_i = server_hello.find(b"\0", 1)
```
davet2001, line 202: Not sure how guarded (a￼gainst bad data) you need the API to be but you may want to handle the cases of zero length message and string that is \0 only.

davet2001 (commented):
Not a full review as I am not familiar with .proto, but what I see looks good. One minor comment.

aioesphomeapi/_frame_helper.py
OttoWinter: Both are already handled - `find` returns -1 when the subsequence is not found. So both zero-length messages and `b"\0"` will return -1

jesserockz (approved):

PR #123, opened 2021-10-21
Dont inherit icon for camera entities

jesserockz:
Fixes an issue caused by #106


jesserockz:
Yeah I did see that, and I was conflicted as to whether camera should support a custom icon. The `config_validation` will not allow an icon to be set by the user, but the underlying code supports it.

paulmonigatti:
Apologies, this was indeed an oversight. esphome/esphome#2600 makes the associated changes on the ESPHome side.

PR #137, opened 2021-11-18
Add system entity category

oxan:
See esphome/esphome#2763.


oxan:
CI error is pre-existing...

jesserockz (approved):

frenck:
`system` should not be used, it is for internal use inside Home Assistant. They are not meant for interaction with the user (and not provided by a user either).

PR #157, opened 2021-12-24
Add device class support to Switch

frenck:
I tried to assign a device class to a switch in ESPHome today... to discover the Switch component doesn't support that? 😬 

First step in adding that.


jesserockz (approved):

jesserockz:
Thanks @frenck

PR #158, opened 2021-12-28
Add lock entity

kbickar:
Adds the lock component to the API

Associated with: 
https://github.com/esphome/esphome/pull/2958


jesserockz:
Duplicate of #142 that I have forgotten about

aioesphomeapi/core.py
```diff
@@ -169,4 +172,7 @@ class ReadFailedAPIError(APIConnectionError):
     57: SirenCommandRequest,
     61: ListEntitiesButtonResponse,
     62: ButtonCommandRequest,
+    63: ListEntitiesLockResponse,
+    64: LockStateResponse,
+    65: LockCommandRequest,
 }
```
jesserockz, line 178: 58, 59 and 60 are already reserved for Lock as per #142 
If you change to those here and in `api.proto` then this PR can be merged.

jesserockz (changes requested):

aioesphomeapi/core.py
```diff
@@ -169,4 +172,7 @@ class ReadFailedAPIError(APIConnectionError):
     57: SirenCommandRequest,
     61: ListEntitiesButtonResponse,
     62: ButtonCommandRequest,
+    63: ListEntitiesLockResponse,
+    64: LockStateResponse,
+    65: LockCommandRequest,
```
jesserockz, line 177: You will need to change the id numbers here too

jesserockz (approved):

PR #232, opened 2022-07-01
climate: add support for quiet fan mode

MichaelMure:
Following https://github.com/esphome/esphome/pull/3609


MichaelMure:
I tried to rebase and regenerate the protobuf files, but for some reason the output is very different than the original file. I followed the instruction in `README.rst`.

MichaelMure:
I managed to regenerate properly those file. I have to say, for someone not familiar with python this is very confusing.

My understanding is that what actually matter is the **system** wide `protoc`, not anything in `requirements.txt`. The docker setup is a good idea but nothing explain how to use it. I had to manually hack in there until it worked.

jesserockz (approved):

PR #255, opened 2022-09-18
Add async-timeout to requirements.txt

mjonuschat:
Based on the history of commits #243 introduced a dependency on
async-timeout but missed adding it to the requirements.txt,
leading to exceptions when trying to use the module in a fresh
environment.


jesserockz (approved):
Thanks, I actually noticed this recently and added it in #259 But will merge this first

PR #268, opened 2022-10-02
Only try remove callbacks when not None

jesserockz:


aioesphomeapi/client.py
```diff
@@ -442,8 +442,8 @@ def on_msg(msg: message.Message) -> None:
         )
 
         def unsub() -> None:
-            assert self._connection is not None
-            self._connection.remove_message_callback(on_msg)
+            if self._connection is not None:
```
bdraco, line 445: I think there are couple of these.

bdraco:
checking to see if there is a place that needs to be cleared on disconnect as well

bdraco (approved):
LGTM, testing now

bdraco:
Testings

- [ ] reload
- [ ] forced disconnect

PR #271, opened 2022-10-06
Fix process_task not ending correctly

jesserockz:


aioesphomeapi/connection.py
```diff
@@ -116,6 +118,12 @@ async def _cleanup(self) -> None:
             await self._frame_helper.close()
             self._frame_helper = None
 
+        if self._process_task is not None:
+            self._process_task.cancel()
+            with suppress(asyncio.CancelledError):
+                await self._process_task
+                self._process_task = None
```
bdraco, line 125:
```suggestion
            with suppress(asyncio.CancelledError):
                await self._process_task
            self._process_task = None
```

The await will likely raise but I think we still want to set it to `None`

jesserockz: Good point

bdraco (approved):
LGTM 👍 

Was able to reload many times without any errors now

PR #272, opened 2022-10-07
Raise GATT errors on read and write etc

jesserockz:
- https://github.com/esphome/esphome/pull/3884


bdraco:
Testing looks good.

Would be nice to show the address as human readable
https://github.com/esphome/aioesphomeapi/pull/291

bdraco (approved):

PR #273, opened 2022-10-07
Fix permanent disconnect issue

### LINKED ISSUE #258: Malformed packet received during handshake results in indefinite pause
**Background:**
One of my devices is in an area with poor network connectivity and frequently disconnects and reconnects. Sometimes (time between 12 hours - 1 month), the device will disconnect and never reconnect even though it is connected to the wifi network.

**Diagnosis:**
After carefully monitoring the issue over a number of months and adding extra debug lines as necessary, it appears that this results from a perfectly timed disconnect during the handshake process.

Some extra logs were added and the relevant logs are displayed below.
```
2022-09-24 02:33:49.069 DEBUG (MainThread) [aioesphomeapi.connection] sonoff_switch_2 @ sonoff_switch_2.local: Got message of type <class 'api_pb2.ConnectResponse'>: 
2022-09-24 02:33:49.070 INFO (MainThread) [aioesphomeapi.connection] sonoff_switch_2 @ sonoff_switch_2.local: Error while reading incoming messages: Error while reading data: 0 bytes read on a total of 1 expected bytes
<-- Report fatal is run here following the error in reading the message -->
2022-09-24 02:33:49.073 DEBUG (MainThread) [aioesphomeapi.connection] connection complete for sonoff_switch_2 @ sonoff_switch_2.local
```

From this, it seems that the [`connect`](https://github.com/esphome/aioesphomeapi/blob/56c7fa35fd86eb1e3fd510c9c2af9b79afc7a88e/aioesphomeapi/connection.py#L278) method is being run and [`login`](https://github.com/esphome/aioesphomeapi/blob/56c7fa35fd86eb1e3fd510c9c2af9b79afc7a88e/aioesphomeapi/connection.py#L278) is called successfully.  However, a packet is received that is malformed and calls the [`_report_fatal_error`](https://github.com/esphome/aioesphomeapi/blob/56c7fa35fd86eb1e3fd510c9c2af9b79afc7a88e/aioesphomeapi/connection.py#L440) prior to the `connect` method returning. This results in the connection being closed while it remains marked as connected in [`reconnect_logic`](https://github.com/esphome/aioesphomeapi/blob/56c7fa35fd86eb1e3fd510c9c2af9b79afc7a88e/aioesphomeapi/reconnect_logic.py#L145). Since the connection was closed and cleaned up, the disconnect callback is never called and thus a reconnect is never attempted again.

**Solution:**
I'm not sure if the malformed packet received during the connection phase should be ignored or passed along to reset the connection post initialization. Either
1) A mutex is added to `_report_fatal_error` to postphone it from excecuting during the connection phase.
2) Simply ignore the error during the connection phase.

For completeness below are additional unmodified logs:
```
2022-09-24 02:33:41.434 DEBUG (MainThread) [aioesphomeapi.connection] Starting connection for sonoff @ sonoff.local
2022-09-24 02:33:41.443 DEBUG (MainThread) [aioesphomeapi.connection] sonoff @ sonoff.local: Connecting to sonoff.local:6053 (AddrInfo(family=<AddressFamily.AF_INET: 2>, type=<SocketKind.SOCK_STREAM: 1>, proto=6, sockaddr=IPv4Sockaddr(address='10.0.0.179', port=6053)))
2022-09-24 02:33:41.479 DEBUG (MainThread) [aioesphomeapi.connection] sonoff.local: Opened socket
2022-09-24 02:33:41.482 DEBUG (MainThread) [aioesphomeapi.connection] sonoff.local: Sending <class 'api_pb2.HelloRequest'>: client_info: "Home Assistant 2022.8.7"
2022-09-24 02:33:41.482 DEBUG (MainThread) [aioesphomeapi.connection] sonoff.local: finished sending
2022-09-24 02:33:41.510 DEBUG (MainThread) [aioesphomeapi.connection] sonoff @ sonoff.local: Got message of type <class 'api_pb2.HelloResponse'>: api_version_major: 1
api_version_minor: 3
server_info: "sonoff (esphome v1.15.3)"

2022-09-24 02:33:41.511 DEBUG (MainThread) [aioesphomeapi.connection] sonoff @ sonoff.local: Successfully connected ('sonoff (esphome v1.15.3)' API=1.3)
2022-09-24 02:33:41.512 DEBUG (MainThread) [aioesphomeapi.connection] sonoff.local: Sending <class 'api_pb2.ConnectRequest'>: password: "REDACTED"

2022-09-24 02:33:41.512 DEBUG (MainThread) [aioesphomeapi.connection] sonoff.local: finished sending
2022-09-24 02:33:49.068 DEBUG (MainThread) [aioesphomeapi.connection] sonoff @ sonoff.local: Got message of type <class 'api_pb2.GetTimeRequest'>: 
2022-09-24 02:33:49.068 DEBUG (MainThread) [aioesphomeapi.connection] sonoff.local: Sending <class 'api_pb2.GetTimeResponse'>: epoch_seconds: 1664004829
2022-09-24 02:33:49.069 DEBUG (MainThread) [aioesphomeapi.connection] sonoff.local: finished sending
2022-09-24 02:33:49.069 DEBUG (MainThread) [aioesphomeapi.connection] sonoff @ sonoff.local: Got message of type <class 'api_pb2.ConnectResponse'>: 
2022-09-24 02:33:49.070 INFO (MainThread) [aioesphomeapi.connection] sonoff @ sonoff.local: Error while reading incoming messages: Error while reading data: 0 bytes read on a total of 1 expected bytes
2022-09-24 02:33:49.070 DEBUG (MainThread) [aioesphomeapi.connection] report fatal: Before cleanup for sonoff @ sonoff.local
2022-09-24 02:33:49.070 DEBUG (MainThread) [aioesphomeapi.connection] _cleanup: before framehelper for sonoff @ sonoff.local
2022-09-24 02:33:49.070 DEBUG (MainThread) [aioesphomeapi.connection] _cleanup: before socket for sonoff @ sonoff.local
2022-09-24 02:33:49.071 DEBUG (MainThread) [aioesphomeapi.connection] _cleanup: before on stop for sonoff @ sonoff.local
2022-09-24 02:33:49.071 DEBUG (MainThread) [aioesphomeapi.connection] _cleanup: finished for sonoff @ sonoff.local
2022-09-24 02:33:49.072 DEBUG (MainThread) [aioesphomeapi.connection] report fatal: after cleanup for sonoff @ sonoff.local
2022-09-24 02:33:49.073 DEBUG (MainThread) [aioesphomeapi.connection] connection complete for sonoff @ sonoff.local
2022-09-24 02:33:49.073 DEBUG (MainThread) [aioesphomeapi.reconnect_logic] _try_connect: after connect sonoff @ sonoff.local
2022-09-24 02:33:49.073 INFO (MainThread) [aioesphomeapi.reconnect_logic] Successfully connected to sonoff @ sonoff.local
2022-09-24 02:33:49.073 DEBUG (MainThread) [aioesphomeapi.reconnect_logic] sonoff @ sonoff.local: aioesphome._try_connect setting connected to false
2022-09-24 02:33:49.073 DEBUG (MainThread) [custom_components.esphome] Starting on_connect for sonoff.local
2022-09-24 02:33:49.074 WARNING (MainThread) [custom_components.esphome] Error getting initial data for sonoff.local: Connection not done for sonoff @ sonoff.local!
Traceback (most recent call last):
  File "/config/custom_components/esphome/__init__.py", line 272, in on_connect
    entry_data.device_info = await cli.device_info()
  File "/config/aioesphomeapi/client.py", line 244, in device_info
    self._check_connected()
  File "/config/aioesphomeapi/client.py", line 235, in _check_connected
    raise APIConnectionError(f"Connection not done for {self._log_name}!")
aioesphomeapi.core.APIConnectionError: Connection not done for sonoff @ sonoff.local!
2022-09-24 02:33:49.076 DEBUG (MainThread) [aioesphomeapi.reconnect_logic] _reconnect_once: waiting for sonoff @ sonoff.local
```

alexyao2015:
fixes #258

### DIFF
--- a/aioesphomeapi/connection.py
+++ b/aioesphomeapi/connection.py
@@ -109,34 +109,40 @@ class APIConnection:
 
         self._process_task: Optional[asyncio.Task[None]] = None
 
+        self._connect_lock: asyncio.Lock = asyncio.Lock()
+
     async def _cleanup(self) -> None:
         """Clean up all resources that have been allocated.
 
         Safe to call multiple times.
         """
-        if self._frame_helper is not None:
-            await self._frame_helper.close()
-            self._frame_helper = None
+        async def _do_cleanup():
+            async with self._connect_lock:
+                if self._frame_helper is not None:
+                    await self._frame_helper.close()
+                    self._frame_helper = None
 
-        if self._process_task is not None:
-            self._process_task.cancel()
-            with suppress(asyncio.CancelledError):
-                await self._process_task
-            self._process_task = None
+                if self._process_task is not None:
+                    self._process_task.cancel()
+                    with suppress(asyncio.CancelledError):
+                        await self._process_task
+                    self._process_task = None
 
-        if self._socket is not None:
-            self._socket.close()
-            self._socket = None
+                if self._socket is not None:
+                    self._socket.close()
+                    self._socket = None
 
-        if not self._on_stop_called and self._connect_complete:
-            # Ensure on_stop is called
-            asyncio.create_task(self.on_stop())
-            self._on_stop_called = True
+                if not self._on_stop_called and self._connect_complete:
+                    # Ensure on_stop is called
+                    asyncio.create_task(self.on_stop())
+                    self._on_stop_called = True
 
-        # Note: we don't explicitly cancel the ping/read task here
-        # That's because if not written right the ping/read task could cancel
-        # themself, effectively ending execution after _cleanup which may be unexpected
-        self._ping_stop_event.set()
+                # Note: we don't explicitly cancel the ping/read task here
+                # That's because if not written right the ping/read task could cancel
+                # themself, effectively ending execution after _cleanup which may be unexpected
+                self._ping_stop_event.set()
+
+        asyncio.create_task(_do_cleanup())
 
     async def _connect_resolve_host(self) -> hr.AddrInfo:
         """Step 1 in connect process: resolve the address."""
@@ -305,19 +311,23 @@ class APIConnection:
             if login:
                 await self.login()
 
-        try:
-            # Allow 2 minutes for connect; this is only as a last measure
-            # to protect from issues if some part of the connect process mistakenly
-            # does not have a timeout
-            async with async_timeout.timeout(120.0):
-                await _do_connect()
-        except Exception:  # pylint: disable=broad-except
-            # Always clean up the connection if an error occured during connect
-            self._connection_state = ConnectionState.CLOSED
-            await self._cleanup()
-            raise
+        # A connection lock must be created to avoid potential issues where
+        # connect has succeeded but not yet returned, followed by a disconnect.
+        # See esphome/aioesphomeapi#258 for more information
+        async with self._connect_lock:
+            try:
+                # Allow 2 minutes for connect; this is only as a last measure
+                # to protect from issues if some part of the connect process mistakenly
+                # does not have a timeout
+                async with async_timeout.timeout(120.0):
+                    await _do_connect()
+            except Exception:  # pylint: disable=broad-except
+                # Always clean up the connection if an error occured during connect
+                self._connection_state = ConnectionState.CLOSED
+                await self._cleanup()
+                raise
 
-        self._connect_complete = True
+            self._connect_complete = True
 
     async def login(self) -> None:
         """Send a login (ConnectRequest) and await the response."""
--- a/aioesphomeapi/connection.py
+++ b/aioesphomeapi/connection.py
@@ -110,13 +110,15 @@ class APIConnection:
         self._process_task: Optional[asyncio.Task[None]] = None
 
         self._connect_lock: asyncio.Lock = asyncio.Lock()
+        self._cleanup_task: Optional[asyncio.Task[None]] = None
 
     async def _cleanup(self) -> None:
         """Clean up all resources that have been allocated.
 
         Safe to call multiple times.
         """
-        async def _do_cleanup():
+
+        async def _do_cleanup() -> None:
             async with self._connect_lock:
                 if self._frame_helper is not None:
                     await self._frame_helper.close()
@@ -142,7 +144,8 @@ class APIConnection:
                 # themself, effectively ending execution after _cleanup which may be unexpected
                 self._ping_stop_event.set()
 
-        asyncio.create_task(_do_cleanup())
+        if not self._cleanup_task or not self._cleanup_task.done():
+            self._cleanup_task = asyncio.create_task(_do_cleanup())
 
     async def _connect_resolve_host(self) -> hr.AddrInfo:
         """Step 1 in connect process: resolve the address."""


aioesphomeapi/connection.py
```diff
@@ -109,34 +109,40 @@ def __init__(
...
+                    # Ensure on_stop is called
+                    asyncio.create_task(self.on_stop())
+                    self._on_stop_called = True
+
+                # Note: we don't explicitly cancel the ping/read task here
+                # That's because if not written right the ping/read task could cancel
+                # themself, effectively ending execution after _cleanup which may be unexpected
+                self._ping_stop_event.set()
+
+        asyncio.create_task(_do_cleanup())
```
bdraco, line 145: It looks like this task could end up still running while it's trying to connect again when it awaits and returns control to the event loop

aioesphomeapi/connection.py
```diff
@@ -305,19 +311,23 @@ async def _do_connect() -> None:
...
-            # Always clean up the connection if an error occured during connect
-            self._connection_state = ConnectionState.CLOSED
-            await self._cleanup()
-            raise
-
-        self._connect_complete = True
+        # A connection lock must be created to avoid potential issues where
+        # connect has succeeded but not yet returned, followed by a disconnect.
+        # See esphome/aioesphomeapi#258 for more information
+        async with self._connect_lock:
```
bdraco, line 317: This makes sense but I think the disconnect needs to run under the lock as well

aioesphomeapi/connection.py
alexyao2015: That's right, but the key is that it doesn't perform disconnect actions until after the initial connect has returned. Specifically, `on_stop` is called after preventing a permanent disconnect scenario.

aioesphomeapi/connection.py
alexyao2015: This shouldn't be necessary since the disconnect isn't doing anything that should be conflicting with the connection process. Aka there should be no additional packets coming in to connect the device while it's disconnecting as was the case with connecting.

aioesphomeapi/connection.py
bdraco: I think it would be a bit cleaner to move the contents of _do_cleanup into a new function called _do_cleanup_locked, hold the lock in _do_cleanup, and call _do_cleanup_locked under the lock and then you can call the _do_cleanup_locked function from the exception handler in the connect without releasing the lock or creating another task

alexyao2015: To be clear, is the main benefit of this to simply avoid creating a new task in the one scenario with an exception case in connect? Am I missing something or are there additional benefits to doing so? It seems that would make things a lot more complicated for seemingly little gain.

alexyao2015: I can look into storing the task and only queuing a new one if there is no existing task, but I would have to guess there's little performance to be gained from this.

bdraco: > To be clear, is the main benefit of this to simply avoid creating a new task in the one scenario with an exception case in connect? Am I missing something or are there additional benefits to doing so? It seems that would make things a lot more complicated for seemingly little gain.

I'm more concerned about future potential races that could happen between when control is returned to the event loop and then the lock is reacquired to execute the cleanup code. 

Since type of races are very difficult to track down so I always prefer a design where we hold the connection lock for the duration of the connection attempt and subsequent cleanup so we don't have analyze every connect and disconnect call to make sure it's race safe

bdraco: Does cherry-picking this commit https://github.com/esphome/aioesphomeapi/commit/79e520d9f6ed56a5a49726d6d96e06a737c633e2 still solve the issue you are seeing?

alexyao2015: That would result in a deadlock because `report_fatal_error` would get called which calls cleanup.

bdraco: For the purposes of answer the above question, please assume that would be refactored to not be a problem

alexyao2015: I'm not sure I understand. Unless there is a major refactor to the code, I'm not sure how it would be fixed to not result in a deadlock. If there is such a major refactor, I don't see the point of reviewing this change. The connect code relies on errors being raised and passed through. These errors are followed the cleanup method

bdraco: The one in `send_message_await_response_complex` that calls `send_message` is a bit unexpected since it will never await the future it creates because if `send_message` raise `SocketAPIError` which means the the `_read_exception_handlers` never get cleaned up.  I think it is not a problem in practice since I don't think we reuse the APIConnection object.

But this code path could definitely use some refactoring as its far too complex to follow

bdraco: I hate to add another flag, but I don't think you wanted to sign up for refactoring this.

Something like this will work and still be race safe https://github.com/esphome/aioesphomeapi/commit/5017794de616de5b6de2af963d5b29a1c5096d40

alexyao2015: I see another call to `cleanup` at line 569 that might also deadlock in this situation. The ultimate solution to this might be a larger refactor of the entire file. I can see wanting to avoid lingering tasks, but this likely can be avoided by having a single task that is cancelled during shutdown.

alexyao2015:
@bdraco I've been testing this PR for a couple of weeks now on my machine and have found it to be stable. I also haven't had anymore of those issues described in #258 so it seems good to be merged pending a discussion about the lingering tasks.

I recently changed it to check for a preexisting task before creating a new one.

bdraco:
While a larger refactor would be nicer, it looks like this will fix the issue at hand.

I'll put this on my production and run it overnight

alexyao2015:
Yeah honestly a full refactor of the entire file would be nice, but this would at least work in the meantime so devices stay connected. The disconnect issue here had been bothering me for months, maybe an entire year before I finally figured this out myself.

bdraco (approved):

bdraco:
Thanks @alexyao2015

bdraco:
Tested overnight and still all good 👍

PR #276, opened 2022-10-11
Add `supports_stop` to `CoverInfo`

amomchilov:
Here's the corresponding change in ESP Home: https://github.com/esphome/esphome/pull/3897
Here's the corresponding change in Home Assistant to consume this value: https://github.com/home-assistant/core/pull/80104


aioesphomeapi/model.py
```diff
@@ -172,7 +173,7 @@ class CoverOperation(APIIntEnum):
     IS_CLOSING = 2
 
 
-@dataclass(frozen=True)
+@dataclass(frozen=True)
```
amomchilov, line 176: Oops, this was an accident. I'll amend it out

jesserockz:
```suggestion
@dataclass(frozen=True)
```

jesserockz (changes requested):
You need to add the new field to the api.proto and regenerate the proto files.

amomchilov:
@jesserockz Hey Jesse.

> You need to add the new field to the api.proto and regenerate the proto files.

Done! Is there anything else I need to do here?

jesserockz (approved):

PR #287, opened 2022-10-24
Adjust ble timeouts to be longer than bleak timeouts

### LINKED ISSUE #286: Timeout in send_message_await_response_complex too short?
Could the timeout timeout: float = 10.0, in send_message_await_response_complex be the too short for ble response? Sometimes it takes a while to read the services and characteristics.

Thank you!

bdraco:
The default timeouts were shorter than the timeout used by bleak and the surrounding stack which could lead to unexplained timeout failures.

fixes #286

### DIFF
--- a/aioesphomeapi/client.py
+++ b/aioesphomeapi/client.py
@@ -153,6 +153,8 @@ from .model import (
 
 _LOGGER = logging.getLogger(__name__)
 
+DEFAULT_BLE_TIMEOUT = 30.0
+
 ExecuteServiceDataType = Dict[
     str, Union[bool, int, float, str, List[bool], List[int], List[float], List[str]]
 ]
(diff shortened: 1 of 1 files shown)


bdraco:
I'm able to successfully pair the eve motion with the adjusted timeouts.

virtualm2000:
Looks very promising. I restarted few times and even if I had to reload the integration for a few devices, once it starts is very stable. Also it recovers very well from the esp32 disconnect problem.


Thank you!

bdraco:
thanks for testing

It looks like we also need to raise bleak error when the device disconnects in the middle of an op but that probably needs to happen on the esphome bleak client instead of here.

bdraco:
This should fix the missing retries https://github.com/home-assistant/core/pull/80885

This should improve the handling of running of connection slots
https://github.com/Bluetooth-Devices/bleak-retry-connector/pull/56

PR #297, opened 2022-10-31
Improve bluetooth error reporting

bdraco:
- Log error when connection timed out but already unsubscribed as its usually caused by a soft reset

- Log the address we were trying to connect to when there is a timeout

- Log a description of BluetoothGATTAPIError

### DIFF
--- a/aioesphomeapi/core.py
+++ b/aioesphomeapi/core.py
@@ -91,9 +91,51 @@ from .api_pb2 import (  # type: ignore
 
 TWO_CHAR = re.compile(r".{2}")
 
-
+# Taken from esp_gatt_status_t in esp_gatt_defs.h
 ESPHOME_GATT_ERRORS = {
-    -1: "Not connected",
+    -1: "Not connected",  # Custom ESPHome error
+    1: "Invalid handle",
+    2: "Read not permitted",
+    3: "Write not permitted",
+    4: "Invalid PDU",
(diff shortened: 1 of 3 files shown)


aioesphomeapi/core.py
```diff
@@ -92,6 +92,12 @@
 TWO_CHAR = re.compile(r".{2}")
 
 
+ESPHOME_GATT_ERRORS = {
+    -1: "Not connected",
+    -2: "Wrong address",
```
bdraco, line 97:
```suggestion
```

I guess we can drop this now based on [esphome/esphome@`f8c13aa` (#3971)](https://github.com/esphome/esphome/pull/3971/commits/f8c13aa9ed7152de1ae22e88a83f8438d642b4a3)

jesserockz (approved):

aioesphomeapi/core.py
```diff
@@ -91,6 +91,53 @@
 
 TWO_CHAR = re.compile(r".{2}")
 
+# Taken from esp_gatt_status_t in esp_gatt_defs.h
+ESPHOME_GATT_ERRORS = {
```
bdraco, line 95: I think we should put this at top level

bdraco: 156e1230d9ec6d4e4619a8440b8aa3dbac86ebc2

jesserockz (approved):

PR #298, opened 2022-11-06
Fix BLE writes timing out

bdraco:
The GATT writes were not using the DEFAULT_BLE_TIMEOUT. When a HomeKit device pairs it can take a while to respond due to the encryption. We need to wait at least 30s

Also logs how long the timeout was when we raise


bdraco:
Similar to https://github.com/hbldh/bleak/issues/926 but just a bit higher since the esps are slower

bdraco:
related https://github.com/home-assistant/core/issues/81416

jesserockz (approved):

Issue #301, opened 2022-11-13
Upgrade protocol buffers


allenporter:
From discussion in https://github.com/home-assistant/core/pull/80432 we need this library to support newer versions of protocol buffer in order to move forward other packages.

I tried to upgrade this myself, but i'm having a hard time understanding the protoc workflow in this library.

```
______________________________ ERROR collecting tests/test_client.py _______________________________
tests/test_client.py:6: in <module>
    from aioesphomeapi.api_pb2 import (
aioesphomeapi/__init__.py:3: in <module>
    from .client import APIClient
aioesphomeapi/client.py:21: in <module>
    from .api_pb2 import (  # type: ignore
aioesphomeapi/api_pb2.py:16: in <module>
    from . import api_options_pb2 as api__options__pb2
aioesphomeapi/api_options_pb2.py:36: in <module>
    _descriptor.EnumValueDescriptor(
venv/lib/python3.9/site-packages/google/protobuf/descriptor.py:755: in __new__
    _message.Message._CheckCalledFromGeneratedFile()
E   TypeError: Descriptors cannot not be created directly.
E   If this call came from a _pb2.py file, your generated code is out of date and must be regenerated with protoc >= 3.19.0.
E   If you cannot immediately regenerate your protos, some other possible workarounds are:
E    1. Downgrade the protobuf package to 3.20.x or lower.
E    2. Set PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python (but this will use pure-Python parsing and will be much slower).
E   
E   More information: https://developers.google.com/protocol-buffers/docs/news/2022-05-06#python-updates
```

PR #307, opened 2022-11-15
Protobuf version upgrades

jesserockz:
I have tested this locally in my dev instance of HA and it is connecting fine to ESPHome devices with state updates etc.

There was no new version of the proto compiler available in the debian repositories so I am now downloading directly from the source releases as advised by https://grpc.io/docs/protoc-installation/

Closes https://github.com/esphome/aioesphomeapi/issues/301


jesserockz:
The `Check protobuf files match` CI will not pass due to changes to the Docker image, but CI uses the published image only.

frenck (commented):
Besides a merge conflict, this looks good to me 👍

PR #316, opened 2022-11-26
Fix connection subscription leak on cancelation

### LINKED ISSUE #315: If the connect task is the unsub is never done and leak
https://github.com/esphome/aioesphomeapi/blob/14a2e30ec379ddeaa488abb509ec761f2bca98c5/aioesphomeapi/client.py#L508

bdraco:
If the connection attempt was canceled the message callback subscription would leak.

fixes #315

### DIFF
--- a/aioesphomeapi/client.py
+++ b/aioesphomeapi/client.py
@@ -518,34 +518,47 @@ class APIClient:
                 self._connection.remove_message_callback(on_msg)
 
         try:
-            async with async_timeout.timeout(timeout):
-                await event.wait()
-        except asyncio.TimeoutError as err:
-            # Disconnect before raising the exception to ensure
-            # the slot is recovered before the timeout is raised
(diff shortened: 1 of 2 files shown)


bdraco:
There is still another leak on cancelation but I can't find it.  Probably will have to dig more tomorrow

aioesphomeapi/client.py
```diff
@@ -518,34 +518,46 @@ def unsub() -> None:
                 self._connection.remove_message_callback(on_msg)
 
         try:
```
bdraco, line 520: Bleak does something similar here https://github.com/hbldh/bleak/blob/master/bleak/backends/bluezdbus/client.py#L159 but catches `BaseException`

I think catching `asyncio.CancelledError` is probably enough here though.

I wasn't able to get it to leak anymore.

jesserockz (approved):

PR #323, opened 2022-11-30
Move message parsing out of the read loop

bdraco:
To further reduce the pressure on the esp, we want to read and drain the buffer as fast as possible, and since we have to wait to read while the data is processing we want the read loop to read as much as possible
first to keep the latency down and reduce the memory pressure on the esp.

### DIFF
--- a/aioesphomeapi/connection.py
+++ b/aioesphomeapi/connection.py
@@ -105,7 +105,7 @@ class APIConnection:
 
         self._ping_stop_event = asyncio.Event()
 
-        self._to_process: asyncio.Queue[message.Message] = asyncio.Queue()
+        self._to_process: asyncio.Queue[Packet] = asyncio.Queue()
 
         self._process_task: Optional[asyncio.Task[None]] = None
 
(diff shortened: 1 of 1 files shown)


aioesphomeapi/connection.py
```diff
@@ -511,46 +511,47 @@ async def _report_fatal_error(self, err: Exception) -> None:
...
-                msg = await self._to_process.get()
+                pkt = await self._to_process.get()
             except RuntimeError:
                 break
 
+            msg_type = pkt.type
+            raw_msg = pkt.data
+            if msg_type not in MESSAGE_TYPE_TO_PROTO:
+                _LOGGER.debug("%s: Skipping message type %s", self.log_name, msg_type)
+                return
```
bdraco, line 529:
```suggestion
                continue
```

bdraco:
I'll move the readexactly change to a new PR.

Went from 25000 to 9000 per minute so nice improvement

bdraco:
the long uuid conversion have a double lower call

Screenshot 2022-11-30 at 8 57 53 AM

bdraco:
Last push is much better
Screenshot 2022-11-30 at 9 06 17 AM

bdraco:
Will move the message parsing optimizations to a new PR as well when I'm done going through all this

jesserockz (approved):

PR #331, opened 2022-12-03
Avoid waiting for process queue when data is already available

bdraco:
We tend to get LE advertisements in burst of up to 16 at a time because of the way the tracker works.  Try to process all of them before returning control to the event loop to reduce latency


bdraco:
This was only 3% faster because the read function returns control to the event loop so often it can't fill up the queue fast enough.

PR #337, opened 2022-12-07
Only start task if existing one is done

alexyao2015:
I think the original code had a small bug.


bdraco:
Pushed to my production.

Hoping it solves the issue I see in https://github.com/home-assistant/core/pull/83888

Will let it run overnight and do a release if its all good

bdraco:
Thanks @alexyao2015

PR #382, opened 2023-02-13
Ensure we hold strong references to tasks

bdraco:
https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task

> Important Save a reference to the result of this function, to avoid a task disappearing mid-execution. The event loop only keeps weak references to tasks. A task that isn’t referenced elsewhere may get garbage collected at any time, even before it’s done. For reliable “fire-and-forget” background tasks, gather them in a collection:

see https://github.com/python/cpython/issues/88831

### DIFF
--- a/aioesphomeapi/connection.py
+++ b/aioesphomeapi/connection.py
@@ -158,7 +158,7 @@ class APIConnection:
 
         if self.on_stop and self._connect_complete:
 
-            def _remove_on_stop_task(_fut: asyncio.Future) -> None:
+            def _remove_on_stop_task(_fut: asyncio.Future[None]) -> None:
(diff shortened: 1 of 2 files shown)


aioesphomeapi/connection.py
```diff
@@ -151,8 +157,19 @@ def _cleanup(self) -> None:
             self._socket = None
 
         if self.on_stop and self._connect_complete:
+
+            def _remove_on_stop_task(_fut: asyncio.Future[None]) -> None:
+                """Remove the stop task from the reconnect loop.
```
bdraco, line 162:
```suggestion
                """Remove the stop task.
```

jesserockz (approved):

PR #386, opened 2023-02-16
Default `Cover.supports_stop` to `True` for backwards compatibility

amomchilov:
Prior to `supports_stop` being sent by ESPHome (introduced in https://github.com/esphome/esphome/pull/3897), all covers were assumed to be support stopping. The whole point of https://github.com/esphome/issues/issues/3059 was to address that.

However, to preserve the correct behaviour for old versions, this PR tweaks `support_stop` (introduced in #276) to default to `True` instead of `False`.

Context: https://github.com/home-assistant/core/pull/80104#discussion_r1106007605

### DIFF
--- a/aioesphomeapi/model.py
+++ b/aioesphomeapi/model.py
@@ -161,7 +161,10 @@ class BinarySensorState(EntityState):
 @dataclass(frozen=True)
 class CoverInfo(EntityInfo):
     assumed_state: bool = False
-    supports_stop: bool = False
+    # "True" by default for backwards compatibility.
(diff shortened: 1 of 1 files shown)


jesserockz:
I have solved this by bumping the api version on the esphome side: [esphome/esphome@`7283c85` (#3897)](https://github.com/esphome/esphome/pull/3897/commits/7283c85febe58a8abf3b95a01f8e641fde0f3bfd)

And then checking the version on the HA side: [home-assistant/core@`3a04927` (#80104)](https://github.com/home-assistant/core/pull/80104/commits/3a04927065c30c813a47fe4f34377986e3056a55)

PR #390, opened 2023-02-22
add pairing methods

jagheterfredrik:
relates to https://github.com/esphome/esphome/pull/4475


bdraco:
We need to be able to give feedback to the API caller if the pairing was successful or not

https://github.com/home-assistant/core/pull/88603/files#diff-972150b31eac2d4abbf3a9e71009b8ed20d32b4ae0c89253fc74d2248d5b6d1dR394

jagheterfredrik:
I'm not sure we do: https://bleak.readthedocs.io/en/latest/api/client.html#bleak.BleakClient.pair

bdraco:
https://github.com/esphome/esphome/pull/4475#pullrequestreview-1309388432

jagheterfredrik:
@bdraco Thoughts? 🙏

bdraco:
Ideally we would get the failure reason code to bubble up so when we get a failure report we don't have to ask someone to dig thorough the logs on their esphome console

https://github.com/jagheterfredrik/esphome/blob/a8e89fb37117e26372dbe242d51386c49f088918/esphome/components/esp32_ble_client/ble_client_base.cpp?rgh-link-date=2023-02-22T15%3A03%3A10Z#L262

jagheterfredrik:
I feel like I might be bloating the BluetoothDeviceConnectionResponse type at that point. Should I just make a new cb for pairing specifically like we discussed in the other PR?

bdraco:
> Should I just make a new cb for pairing specifically like we discussed in the other PR?

Yes. Most devices won't need it so it seems like it should be it's own thing

bdraco:
LGTM. Will need Jesse's feedback on the esphome PR before we can merge this.

jagheterfredrik:
Thanks @bdraco, I'll clean up and test

jagheterfredrik:
The proto did not compile as cleanly here as it did for esphome, not sure if it's using some older protoc version?

bdraco:
> The proto did not compile as cleanly here as it did for esphome, not sure if it's using some older protoc version?

Use the docker command in the GitHub actions file

jagheterfredrik:
Ah ofc cheers

bdraco (approved):
Should be good to merge once the esphome PR is approved. This may need to change a bit based on feedback from Jesse. He is on the other side of the dateline so probably will be a few days before you get feedback there.

jesserockz (approved):

PR #391, opened 2023-02-23
Fix bluetooth_device_connect returning prematurely from a disconnect

jagheterfredrik:
fixes an issue when a disconnect is followed by a connect, example:

```python
def connect_cb(connected, ...)
   print("Connected:", connected)
await api.bluetooth_device_disconnect(...)
await api.bluetooth_device_connect(..., connect_cb)
print('Attempt to use connection')
```
yields
```
Connected: False
Attempt to use connection
Connected: True
```
`bluetooth_device_disconnect` resolves as soon as the aioesphomeapi call is done which I think is fine but it is unexpected to have a closed connection after an `await bluetooth_device_connect`.

### DIFF
--- a/aioesphomeapi/client.py
+++ b/aioesphomeapi/client.py
@@ -497,7 +497,8 @@ class APIClient:
             resp = BluetoothDeviceConnection.from_pb(msg)
             if address == resp.address:
                 on_bluetooth_connection_state(resp.connected, resp.mtu, resp.error)
-                event.set()
+                if resp.connected:
+                    event.set()
 
         assert self._connection is not None
         if has_cache:


bdraco (approved):

jagheterfredrik:
Note that HA does await connected status https://github.com/home-assistant/core/blob/dev/homeassistant/components/esphome/bluetooth/client.py#L335 so not sure if this is... useful

bdraco:
> Note that HA does await connected status [home-assistant/core@`3011449`/homeassistant/components/esphome/bluetooth/client.py#L335](https://github.com/home-assistant/core/blob/301144993cb39d9dc8c4fd752d7d50c3fc9f218d/homeassistant/components/esphome/bluetooth/client.py#L335) so not sure if this is... useful. I guess it could simplify the HA code.

Your fix is still helpful to HA because we would still not run the exception handler if the connect actually fails.

bdraco:
I don't expect any side effects, but I'll run this overnight on my production since it connects and disconnects often.

If its all good, I'll do a release with this fix tomorrow and bump HA

jagheterfredrik:
Cool, thanks

bdraco:
Thanks @jagheterfredrik

PR #393, opened 2023-03-02
If a field value is None, set the value to default before passing it to a converter

ov1d1u:
This is related to https://github.com/home-assistant/core/issues/88887

Seems that there is no check for `None` values before calling a field's converter function, so this might fails for some specific use cases. To fix that, in `APIModelBase`'s `__post_init__` I'm setting the value to the field's default value before calling the converter. I'm not familiar with this project so I'm not sure if this was the purpose of `field_.default`, feel free to discard this PR if I misunderstood that.

### DIFF
--- a/aioesphomeapi/model.py
+++ b/aioesphomeapi/model.py
@@ -69,6 +69,8 @@ class APIModelBase:
             if convert is None:
                 continue
             val = getattr(self, field_.name)
+            if val is None:
+                val = field_.default
             # use this setattr to prevent FrozenInstanceError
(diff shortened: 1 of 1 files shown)


jesserockz:
So I followed the chain of linked issues etc. The actual issue here is that the custom esphome component for opentherm is not defining a `step` which is a required field in ESPHome `number` components. You need to report an issue there and get it fixed.

ov1d1u:
Thank you for clarifying this. I also created an issue on the component's repo, and even a PR. As a result, I'm now closing this since I guess it's no longer needed.

PR #339, opened 2022-12-13
Fix failure to reconnect when the process task raises an exception during decoding a protobuf message

bdraco:
When the process loop raised an exception it would bubble through into the cleanup
which would prevent the disconnect callback from being fired.  In this case it
was UnicodeDecodeError from a bad name in advertisment.   We now handle
unicode decode errors by replacing invalid utf-8 and catch the exception
so it does not prevent the disconnected callback from running

seen https://github.com/home-assistant/core/pull/83888#issuecomment-1347670905

PR #408, opened 2023-03-25
update to latest api.proto

richardhopton:


aioesphomeapi/api.proto
```diff
@@ -283,7 +284,6 @@ message ListEntitiesCoverResponse {
   bool disabled_by_default = 9;
   string icon = 10;
   EntityCategory entity_category = 11;
-  bool supports_stop = 12;
```
richardhopton, line 286: Perhaps this should be kept with `[deprecated = true]` or something

aioesphomeapi/api.proto
```diff
@@ -610,7 +623,7 @@ message SubscribeLogsResponse {
   option (no_delay) = false;
 
   LogLevel level = 1;
-  bytes message = 3;
+  string message = 3;
```
richardhopton, line 626: This is a breaking change - but super convenient having the name coerced to a string.

aioesphomeapi/api.proto
```diff
@@ -1166,12 +1183,13 @@ message BluetoothLEAdvertisementResponse {
   option (no_delay) = true;
 
   uint64 address = 1;
-  bytes name = 2;
+  string name = 2;
```
richardhopton, line 1186: This is another breaking change - but super convenient having the name coerced to a string.

richardhopton:
I didn't remove the siren stuff that @jesserockz added to this repo but not to esphome repo.

aioesphomeapi/api.proto
jesserockz: Why are you removing this field? It was only just added recently

aioesphomeapi/api.proto
jesserockz: https://github.com/esphome/aioesphomeapi/pull/339

aioesphomeapi/api.proto
richardhopton: I copied the api.proto from here: https://github.com/esphome/esphome/blob/dev/esphome/components/api/api.proto
and these were all the things that changed... I was clearly incorrect to think that keeping these two files in sync was a good idea.

PR #410, opened 2023-03-26
Add the ability to clear the cache to the API

bdraco:
With BlueZ we can fully clear the cache when something goes wrong with the services, however since this is also a cache on the ESP32 we need to be able to clear the on device cache as well for the proxies since if something goes wrong with the service resolution it can cache the bad resolution on NVS forever.

Our current client implementation is limited to clearing the memory cache in Home Assistant https://github.com/home-assistant/core/blob/89355e087952417a6824507fd3b197f9d0520e19/homeassistant/components/esphome/bluetooth/client.py#L512

related issue https://github.com/esphome/issues/issues/4156

https://github.com/esphome/esphome/pull/4621


jesserockz (changes requested):
Need to rebase and bump the message id as there was another PR that added an unsub message.

Also, need to add the class and id to the `MESSAGE_TYPE_TO_PROTO` in `core.py`

aioesphomeapi/api.proto
```diff
@@ -1392,3 +1393,13 @@ message UnsubscribeBluetoothLEAdvertisementsRequest {
   option (id) = 87;
   option (source) = SOURCE_CLIENT;
 }
+
+message BluetoothDeviceClearCacheResponse {
+  option (id) = 89;
```
jesserockz, line 1398:
```suggestion
  option (id) = 88;
```

jesserockz (approved):

bdraco:
Thanks

PR #412, opened 2023-04-03
Add initial voice assistant support

jesserockz:
Related PRs:

- https://github.com/home-assistant/core/pull/90691
- https://github.com/esphome/esphome/pull/4648


aioesphomeapi/client.py
```diff
@@ -1238,3 +1245,81 @@ def api_version(self) -> Optional[APIVersion]:
...
+                    self._connection.send_message(VoiceAssistantResponse(error=True))
+
+        def on_msg(msg: VoiceAssistantRequest) -> None:
+            command = VoiceAssistantCommand.from_pb(msg)
+            loop = asyncio.get_running_loop()
+            if command.start:
+                t = loop.create_task(handle_start())
+                t.add_done_callback(_started)
+            else:
+                loop.create_task(handle_stop())
```
bdraco, line 1283: We should hold a reference to these so they do not get garbage collected mid execution

https://docs.python.org/3/library/asyncio-task.html#creating-tasks

From the asyncio tasks section:
> Important Save a reference to the result of this function, to avoid a task disappearing mid-execution. The event loop only keeps weak references to tasks. A task that isn’t referenced elsewhere may get garbage collected at any time, even before it’s done. For reliable “fire-and-forget” background tasks, gather them in a collection:

bdraco:
Just realized I never actually saved the comment above. 🙈

PR #424, opened 2023-04-20
Hold strong references to voice assistant tasks

bdraco:
`start_task` has a reference in `unsub` which the caller holds
`stop_task` did not have a strong reference. To prevent it from being garbage collected its now held in background tasks

https://github.com/esphome/aioesphomeapi/pull/412/files#r1162523622


bdraco:
I don't have a way to test this

jesserockz (approved):

jesserockz (approved):

PR #428, opened 2023-05-04
Fix disconnecting while handshake is in process

bdraco:
related issue https://github.com/home-assistant/core/issues/92432


bdraco:
I'm able to reload the config entry over and over again without issue now

freddiemare:
thanks bdraco I will test your fix

PR #429, opened 2023-05-07
Fix BLE connection timing out on failure instead of raising

bdraco:
https://github.com/esphome/aioesphomeapi/pull/391 changed the connection callback to only resolve when the connection was successful to work around a race where we get a late disconnected event while connecting. The net result is we now have to wait for the whole timeout on error.

This solution reverts that change and makes disconnect wait up to 5 seconds to try to avoid the condition from happening in the first place.  Previously disconnect was fire and forget which is why the race could happen in the first place.

https://github.com/home-assistant/core/issues/90265

### DIFF
--- a/aioesphomeapi/client.py
+++ b/aioesphomeapi/client.py
@@ -255,6 +255,11 @@ class APIClient:
             return f"{self._cached_name} @ {self.address}"
         return self.address
 
+    def set_cached_name_if_unset(self, name: str) -> None:
+        """Set the cached name of the device if not set."""
+        if not self._cached_name:
+            self._cached_name = name
+
     async def connect(
         self,
         on_stop: Optional[Callable[[bool], Awaitable[None]]] = None,
--- a/aioesphomeapi/client.py
+++ b/aioesphomeapi/client.py
@@ -529,11 +529,8 @@ class APIClient:
             resp = BluetoothDeviceConnection.from_pb(msg)
 
             if address == resp.address:
-
-                _LOGGER.warning("on_bluetooth_connection_state address=%s resp=%s", address, resp)
                 on_bluetooth_connection_state(resp.connected, resp.mtu, resp.error)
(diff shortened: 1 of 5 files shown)


bdraco:
This regressed in https://github.com/esphome/aioesphomeapi/pull/391

aioesphomeapi/client.py
```diff
@@ -527,10 +527,10 @@ async def bluetooth_device_connect(  # pylint: disable=too-many-locals
 
         def on_msg(msg: BluetoothDeviceConnectionResponse) -> None:
             resp = BluetoothDeviceConnection.from_pb(msg)
+
```
bdraco, line 530:
```suggestion
```

bdraco:
2023-05-07 13:58:32.155 WARNING (MainThread) [aioesphomeapi.client] 98:1B:B5:69:0D:59: Bluetooth device disconnected after 0.0006275790219660848s

works as expected

aioesphomeapi/client.py
```diff
@@ -693,15 +696,24 @@ def predicate_func(msg: BluetoothDeviceClearCacheResponse) -> bool:
...
+        self, address: int, timeout: float = DEFAULT_BLE_DISCONNECT_TIMEOUT
+    ) -> None:
         self._check_authenticated()
 
+        def predicate_func(msg: BluetoothDeviceConnectionResponse) -> bool:
+            return bool(msg.address == address and not msg.connected)
+
         assert self._connection is not None
-        self._connection.send_message(
+        await self._connection.send_message_await_response_complex(
```
bdraco, line 708: Waiting is fine because we always reply if already disconnected

https://github.com/esphome/esphome/blob/3498aade85f10e342efcd03bccc1dcdeceff5b5f/esphome/components/bluetooth_proxy/bluetooth_proxy.cpp#L272

PR #435, opened 2023-05-13
Reduce overhead to process bluetooth advertisements

bdraco:
Since we only ever convert these from protobuf we can reduce the branching and collapse the conversion functions.

The bytes recopy is now avoided in the dict comps.

Will run this on production for a while to be sure

Processing these is still the largest time for all the esphome code.

Screenshot 2023-05-13 at 4 27 34 PM

This works out to ~17% speed up

### DIFF
--- a/aioesphomeapi/model.py
+++ b/aioesphomeapi/model.py
@@ -6,7 +6,6 @@ from typing import (
     TYPE_CHECKING,
     Any,
     Callable,
-    Collection,
     Dict,
     Iterable,
     List,
@@ -30,7 +29,6 @@ else:
 if TYPE_CHECKING:
     from .api_pb2 import (  # type: ignore
         BluetoothLEAdvertisementResponse,
-        BluetoothServiceData,
         HomeassistantServiceMap,
     )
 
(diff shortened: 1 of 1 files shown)


aioesphomeapi/model.py
```diff
@@ -864,16 +813,46 @@ class BluetoothLEAdvertisement:
     def from_pb(  # type: ignore[misc]
         cls: "BluetoothLEAdvertisement", data: "BluetoothLEAdvertisementResponse"
     ) -> "BluetoothLEAdvertisement":
+        _uuid_convert = _cached_uuid_converter
+
+        if raw_manufacturer_data := data.manufacturer_data:
```
bdraco, line 818: Need to make sure we can use walrus with the minimum python version

aioesphomeapi/model.py
```diff
@@ -864,16 +813,46 @@ class BluetoothLEAdvertisement:
...
+            else:
+                # Legacy data
+                manufacturer_data = {
+                    int(v.uuid, 16): bytes(v.legacy_data) for v in raw_manufacturer_data
+                }
+        else:
+            manufacturer_data = {}
+
+        if raw_service_data := data.service_data:
+            if raw_service_data.data:
```
bdraco, line 832: Need to check element 0

bdraco, line 832:
```suggestion
            if raw_service_data[0].data:
```

aioesphomeapi/model.py
```diff
@@ -801,55 +799,6 @@ def _uuid_converter(uuid: str) -> str:
...
-        _cached_uuid_converter(v.uuid): bytes(v.legacy_data)  # type: ignore[union-attr]
-        for v in value
-    }
-
-
-def _convert_bluetooth_le_manufacturer_data(
-    value: Union[Dict[int, bytes], Collection["BluetoothServiceData"]],
-) -> Dict[int, bytes]:
-    # We literally mean is a `dict` not any other class
-    if type(value) is dict:  # pylint: disable=unidiomatic-typecheck
```
bdraco, line 840: This was never reachable since we only convert from pb

aioesphomeapi/model.py
```diff
@@ -801,55 +799,6 @@ def _uuid_converter(uuid: str) -> str:
...
-        return []
-
-    return [_cached_uuid_converter(v) for v in value]
-
-
-def _convert_bluetooth_le_service_data(
-    value: Union[Dict[str, bytes], Collection["BluetoothServiceData"]],
-) -> Dict[str, bytes]:
-    # We literally mean is a `dict` not any other class
-    if type(value) is dict:  # pylint: disable=unidiomatic-typecheck
```
bdraco, line 817: This was never reachable since we only convert from pb

bdraco:
It’s fast enough now that I no longer get the timeout on the first unlock attempt and than connect again

PR #447, opened 2023-06-24
Fix more cases where incorrect encryption keys were not detected

bdraco:
see https://github.com/home-assistant/core/issues/94663

- Fix encryption error being overridden by another exception

- Fix encryption error not being raised because cleanup fails with a different exception which overshadows it

- Ensure that an exception during an asyncio.Protocol callback is reported upward by switching from an event to a future

- Backoff to maximum time on password or encryption error as we do not want to keep hammering the device if the reason we cannot connect is auth

### DIFF
--- a/aioesphomeapi/_frame_helper.py
+++ b/aioesphomeapi/_frame_helper.py
@@ -9,6 +9,7 @@ import async_timeout
 from noise.connection import NoiseConnection  # type: ignore
 
 from .core import (
+    APIConnectionError,
     BadNameAPIError,
     HandshakeAPIError,
     InvalidEncryptionKeyAPIError,
@@ -16,7 +17,6 @@ from .core import (
     RequiresEncryptionAPIError,
     SocketAPIError,
     SocketClosedAPIError,
-    APIConnectionError,
 )
 from .util import bytes_to_varuint, varuint_to_bytes
 
--- a/aioesphomeapi/_frame_helper.py
+++ b/aioesphomeapi/_frame_helper.py
@@ -264,8 +264,6 @@ class APINoiseFrameHelper(APIFrameHelper):
                 ]
             )
             self._transport.write(header + frame)
-            if debug:
-                _LOGGER.debug("Sending encrypted frame: [%s]", (header + frame).hex())
         except (RuntimeError, ConnectionResetError, OSError) as err:
             raise SocketAPIError(f"Error while writing data: {err}") from err
 
--- a/aioesphomeapi/_frame_helper.py
+++ b/aioesphomeapi/_frame_helper.py
@@ -231,12 +231,14 @@ class APINoiseFrameHelper(APIFrameHelper):
         # Make sure we set the ready event if its not already set
         # so that we don't block forever on the ready event if we
         # are waiting for the handshake to complete.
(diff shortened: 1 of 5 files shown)


aioesphomeapi/_frame_helper.py
```diff
@@ -219,21 +220,29 @@ def __init__(
...
         # Make sure we set the ready event if its not already set
         # so that we don't block forever on the ready event if we
         # are waiting for the handshake to complete.
-        self._ready_event.set()
+        self._set_ready_future_exception(APIConnectionError("Connection closed"))
         self._state = NoiseConnectionState.CLOSED
         super().close()
 
+    def _handle_error_and_close(self, exc: Exception) -> None:
+        self._set_ready_future_exception(exc)
```
bdraco, line 243: This will ensure the exception in the protocol callback raises upwards during the handshake

tests/test__frame_helper.py
```diff
@@ -63,3 +64,41 @@ def _on_error(exc: Exception):
 
         assert type_ == pkt_type
         assert data == pkt_data
+
+
+@pytest.mark.asyncio
+async def test_noise_frame_helper_incorrect_key():
+    """Test that the noise frame helper can perform a handshake with the ESPHome device."""
```
bdraco, line 71:
```suggestion
    """Test that the noise frame helper raises InvalidEncryptionKeyAPIError on bad key."""
```

aioesphomeapi/_frame_helper.py
```diff
@@ -252,6 +264,8 @@ def _write_frame(self, frame: bytes) -> None:
                 ]
             )
             self._transport.write(header + frame)
+            if debug:
+                _LOGGER.debug("Sending encrypted frame: [%s]", (header + frame).hex())
```
bdraco, line 268:
```suggestion
```

aioesphomeapi/_frame_helper.py
```diff
@@ -219,29 +221,39 @@ def __init__(
...
     def _write_frame(self, frame: bytes) -> None:
         """Write a packet to the socket, the caller should not have the lock.
 
         The entire packet must be written in a single call to write
         to avoid locking.
         """
-        _LOGGER.debug("Sending frame %s", frame.hex())
+        debug = _LOGGER.isEnabledFor(logging.DEBUG)
         assert self._transport is not None, "Transport is not set"
+        if debug:
```
bdraco, line 255:
```suggestion
        if _LOGGER.isEnabledFor(logging.DEBUG):
```

aioesphomeapi/_frame_helper.py
```diff
@@ -219,29 +221,39 @@ def __init__(
...
+        super()._handle_error_and_close(exc)
+
     def _write_frame(self, frame: bytes) -> None:
         """Write a packet to the socket, the caller should not have the lock.
 
         The entire packet must be written in a single call to write
         to avoid locking.
         """
-        _LOGGER.debug("Sending frame %s", frame.hex())
+        debug = _LOGGER.isEnabledFor(logging.DEBUG)
```
bdraco, line 253:
```suggestion
```

PR #448, opened 2023-06-25
Propagate if disconnect was expected to disconnect callback

bdraco:
BREAKING CHANGE: The signature for the `on_disconnect` callback has changed to `Callable[[bool], Awaitable[None]]`

The expected_disconnect parameter is used to indicate if the disconnect was expected
or not. This is used to determine if we should reconnect immediately or wait a bit.

For consumers of this class, the expected_disconnect parameter is useful to determine
if a deep sleep device went to sleep or if the connection was lost unexpectedly.


aioesphomeapi/reconnect_logic.py
```diff
@@ -48,7 +51,7 @@ def __init__(
         self._cli = client
         self.name = name
         self._on_connect_cb = on_connect
-        self._on_disconnect_cb = on_disconnect
+        self._on_disconnect_cb = self._make_disconnect_cb(on_disconnect)
```
balloob, line 54: Why do we need to be backwards compatible? We can just note a breaking change in the release notes and update it in HA when we update the lib ?

balloob (approved):

aioesphomeapi/reconnect_logic.py
bdraco: I guess I can adjust the cli as well at the same time

bdraco: I'll take out the back compat and do a pr to esphome or as well

PR #451, opened 2023-06-30
Speed up noise by using chacha20poly1305-reuseable

bdraco:
Using noise currently incurs a performance penalty (enough that not using it is sometimes a understandable reason). It's a bit uncomfortable being in a position of having to ask users to turn off noise because they have so many esphome devices that it's overwhelming their ha running on raspberry pi (eof errors, connection drops, etc). The goal is to get that penalty closer to zero to eliminate any reason not to use noise (if only eap8266 had just a bit more ram we ... but that's a another issue).

chacha20poly1305-reuseable is a faster drop-in replacement version of chacha20poly1305 which does not use the thread locking that the cryptography base implementation uses since asyncio does not need the ability to encrypt/decrypt in different threads.

Home Assistant currently uses `chacha20poly1305-reuseable` for `HAP-python`, `aiohomekit`, and `pyatv` as they also use long running data streams with similar applications

Wheels are available via ciwheelbuild https://pypi.org/project/chacha20poly1305-reuseable/#files and HA is already building wheels for this library https://wheels.home-assistant.io/musllinux/. The library uses cython to reduce python overhead as well but falls back to pure-python if cython is not available. This allows us to avoid a breaking change in the event for some reason the library can't be compiled with cython (the wheels are more widely available than cryptography so this is unlikely). Additionally, the library is well covered https://app.codecov.io/gh/bdraco/chacha20poly1305-reuseable/blob/main/src%2Fchacha20poly1305_reuseable%2F__init__.py and works on python as old as 3.7
 
In testing bluetooth workloads this was ~68-82% less run time with the cython backend, YMMV with architecture as the savings were on the higher side for x86_64 but on the lower side for armv7

### DIFF
--- a/aioesphomeapi/_frame_helper.py
+++ b/aioesphomeapi/_frame_helper.py
@@ -8,9 +8,9 @@ from typing import Any, Callable, Optional, Union, cast
 import async_timeout
 from chacha20poly1305_reuseable import ChaCha20Poly1305Reusable
 from cryptography.exceptions import InvalidTag
-from noise.backends.default import DefaultNoiseBackend
-from noise.backends.default.ciphers import ChaCha20Cipher
-from noise.connection import NoiseConnection  # type: ignore
+from noise.backends.default import DefaultNoiseBackend,  # type: ignore[import]
+from noise.backends.default.ciphers import ChaCha20Cipher # type: ignore[import]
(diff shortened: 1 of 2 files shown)


bdraco:
double checked this is ok with esphome itself + HA on 3 production installs (not that I was expecting anything different here)

PR #454, opened 2023-07-01
Reduce protocol overhead

bdraco:
- Unpack protocol bytes into named values: Easier to see what the data is and uses unpacking sequence which is fast
- For small packets we avoid calling `bytes_to_varuint`
- Avoid a lot of memory copy
- Avoid multiple calls to get length in places it cannot change
- Decryption errors are now raised instead of causing the protocol to fail (its still going to drop but we get the right exception now)

Works out to ~24-38% faster on average


bdraco:
This has been solid on all my production so I'm comfortable moving it foward.

I'll get it bumped in HASS but not tagged for beta so we have a few weeks of it running just in case I missed something (hopefully not since we have a lot better coverage for this now)

PR #456, opened 2023-07-03
Use memoryview for camera images

bdraco:
Save the camera data in a list of memory views and join them all together when we have the final packet. This avoids creating a new bytes object every time the camera sends a partial response


bdraco:
There aren't any tests for the camera platform in hass so the first step was to buy a camera to get a good test case

I noticed all the copy in the process.  Will work on ha side tests next

PR #457, opened 2023-07-03
Use a finally block instead of catching exceptions in send_message_callback_response

bdraco:


bdraco:
nevermind. this isn't going to work

PR #459, opened 2023-07-04
Add slots to all models when using py3.10 or later

bdraco:
All the models are now slotted to reduce memory and overhead.

This would be a lot cleaner for mypy if we could drop py < 3.10 but its easy enough to remove the ignores when we do.


bdraco:
Sadly this won't work until we drop < 3.10

bdraco:
I didn't think I was going to be able to get this to work but testing looks good now

bdraco:
Memory on my big install is finally sitting below 500MiB with this change 👍

bdraco:
Need to test a camera

bdraco:
We can drop all the ignores when we require python 3.10 or later

bdraco:
camera still works great

bdraco:
This is saving almost ~21MiB on my big install.

bdraco:
The construction is ~3% slower

The attribute accesses are ~10% faster

Since we construct these a lot it looks like a performance neutral change (or at least below the margin of error measurement). The memory use reduction is awesome, and given how much of the advertisment data for bluetooth has to sit in memory all the time thats a win just on its own without even considering all the state and static info

PR #461, opened 2023-07-09
Improve performance reassembling fragmented packets

bdraco:
The camera packets tend to get fragmented because of the size and I was looking for a bug in the reassembly but it turns out the camera was just overheating.

I thought there was a bug here but there isn't after I wrote the tests it shows it works as expected. However the new design is faster so I'll clean it up and move it forward

- Avoid checking `len()` each read call
- Remove some unreachable checks since the decryption will fail if the data is invalid
- Remove a useless shift
- Avoid the extra stack depth of `_callback_packet` since it was only called from one function

### DIFF
--- a/aioesphomeapi/_frame_helper.py
+++ b/aioesphomeapi/_frame_helper.py
@@ -125,8 +125,9 @@ class APIPlaintextFrameHelper(APIFrameHelper):
 
     def _callback_packet(self, type_: int, data: Union[bytes, bytearray]) -> None:
         """Complete reading a packet from the buffer."""
-        del self._buffer[: self._pos]
-        self._buffer_len -= self._pos
+        end_of_frame_pos = self._pos
+        del self._buffer[: end_of_frame_pos]
+        self._buffer_len -= end_of_frame_pos
         self._on_pkt(type_, data)
 
     def write_packet(self, type_: int, data: bytes) -> None:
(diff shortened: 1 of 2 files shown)


aioesphomeapi/_frame_helper.py
```diff
@@ -489,18 +489,8 @@ def _handle_frame(self, frame: bytearray) -> None:
                 ProtocolAPIError(f"Bad encryption frame: {ex!r}")
             )
             return
-        msg_len = len(msg)
-        if msg_len < 4:
-            self._handle_error_and_close(ProtocolAPIError(f"Bad packet frame: {msg!r}"))
-            return
```
bdraco, line 495: This should be impossible since the encryption would fail to decrypt if the packet was invalid

aioesphomeapi/_frame_helper.py
```diff
@@ -489,18 +489,8 @@ def _handle_frame(self, frame: bytearray) -> None:
...
-        msg_type_high, msg_type_low, data_len_high, data_len_low = msg[:4]
+        msg_type_high, msg_type_low = msg[:2]
         msg_type = (msg_type_high << 8) | msg_type_low
-        data_len = (data_len_high << 8) | data_len_low
-        if data_len + 4 != msg_len:
-            self._handle_error_and_close(
-                ProtocolAPIError(f"Bad data len: {data_len} vs {msg_len}")
-            )
-            return
```
bdraco, line 503: This should be impossible since the encryption would fail to decrypt if the packet was invalid

PR #463, opened 2023-07-10
Avoid sending pings when we have traffic

bdraco:
This is the same optimization that the ESPHome side has

https://github.com/esphome/esphome/blob/ddde1ee31ef96476c7a7d1c8b8fe306aa4d2e503/esphome/components/api/api_connection.cpp#L119

The goal is to keep network traffic to a minimum and avoid radio time when not needed

### DIFF
--- a/aioesphomeapi/connection.py
+++ b/aioesphomeapi/connection.py
@@ -135,6 +135,8 @@ class APIConnection:
         "_connect_task",
         "_fatal_exception",
         "_expected_disconnect",
+        "_loop",
+        "_send_pending_ping",
     )
 
     def __init__(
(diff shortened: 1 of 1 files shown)


bdraco:
Killed power and made sure the devices still go unavailable and recover

aioesphomeapi/connection.py
```diff
@@ -172,6 +174,8 @@ def __init__(
         self._connect_task: Optional[asyncio.Task[None]] = None
         self._fatal_exception: Optional[Exception] = None
         self._expected_disconnect = False
+        self._send_pending_ping = False
+        self._loop = asyncio.get_event_loop()
```
bdraco, line 178: We can change this to `get_running_loop()` once we drop support for older python

PR #468, opened 2023-07-14
Default unit of measurement to None

joostlek:
As reported in https://github.com/home-assistant/core/issues/96543, the device class for AQI should be None instead of an empty string.

### DIFF
--- a/aioesphomeapi/model.py
+++ b/aioesphomeapi/model.py
@@ -615,7 +615,7 @@ class NumberInfo(EntityInfo):
     step: float = converter_field(
         default=0.0, converter=fix_float_single_double_conversion
     )
-    unit_of_measurement: str = ""
+    unit_of_measurement: Optional[str] = None
     mode: Optional[NumberMode] = converter_field(
         default=NumberMode.AUTO, converter=NumberMode.convert
     )
--- a/aioesphomeapi/model.py
+++ b/aioesphomeapi/model.py
@@ -408,7 +408,7 @@ class LastResetType(APIIntEnum):
 @dataclass(frozen=True)
 class SensorInfo(EntityInfo):
     device_class: str = ""
-    unit_of_measurement: str = ""
+    unit_of_measurement: Optional[str] = None
     accuracy_decimals: int = 0
     force_update: bool = False
     state_class: Optional[SensorStateClass] = converter_field(


aioesphomeapi/model.py
```diff
@@ -408,7 +408,7 @@ class LastResetType(APIIntEnum):
 @dataclass(frozen=True)
 class SensorInfo(EntityInfo):
     device_class: str = ""
-    unit_of_measurement: str = ""
```
bdraco, line 411: https://esphome.io/components/sensor/index.html
> device_class (Optional, string): The device class for the sensor. See https://developers.home-assistant.io/docs/core/entity/sensor/#available-device-classes for a list of available options. Set to "" to remove the default device class of a sensor.


It looks like the docs have the above so I'm not sure this is going to solve the problem

joostlek: In that case, for AQI the default isnt set to None, where should that happen?

bdraco: I think it will fix this specific issue but I'm worried about the side effects of this change since it could mean everyone who hasn't set a device class will now get None instead of "" 

How that is going to affect their setups is not clear here. Best case nobody notices, but it could be a significant breaking change for some.  

It's difficult to understand the consequences of this change without looking through how core and frontend handle the difference between "" and None

bdraco: For clarity, this seems like a good change, but we need to better understand the impact before moving it forward

joostlek: Agreed, I also don't know the internals, but at least it's an attempt :)

bdraco: It looks like the default was added in https://github.com/esphome/aioesphomeapi/commit/10724ea3e54f7a9cd7734b0183ba0e3e25807588

bdraco: related issue https://github.com/home-assistant/core/issues/28157

bdraco: So I don't think this will work since protobuf doesn't serialize nulls.  They need to be explicitly set as another field.

We would have to do something like
```
--- a/aioesphomeapi/api.proto
+++ b/aioesphomeapi/api.proto
@@ -259,6 +259,7 @@ message ListEntitiesBinarySensorResponse {
   bool disabled_by_default = 7;
   string icon = 8;
   EntityCategory entity_category = 9;
+  bool has_device_class = 10;
 }
 message BinarySensorStateResponse {
   option (id) = 21;

```

But if we never need `""` to be sent as a device class we can convert them to `None` instead explicitly in the integration

bdraco: >  the device class for AQI should be None instead of an empty string.

I just realized  this is actually about `unit_of_measurement`

bdraco: I think https://github.com/home-assistant/core/pull/96556 will solve this

Let me know what you think.

PR #469, opened 2023-07-15
Guard a few frequently called instance checks with if TYPE_CHECKING

bdraco:
There were some isinstance checks that can never be false

### DIFF
--- a/aioesphomeapi/client.py
+++ b/aioesphomeapi/client.py
@@ -482,7 +482,9 @@ class APIClient:
         assert self._connection is not None
 
         def is_response(msg: message.Message) -> bool:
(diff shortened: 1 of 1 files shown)


aioesphomeapi/client.py
```diff
@@ -482,11 +482,9 @@ async def _send_bluetooth_message_await_response(
         assert self._connection is not None
 
         def is_response(msg: message.Message) -> bool:
```
bdraco, line 484: This could be a type var instead when we drop old python

bdraco:
Trying to solve https://github.com/esphome/aioesphomeapi/issues/470

Issue #470, opened 2023-07-15
send message complex is too slow


bdraco:
The current client needs to create two closures for each Bluetooth message. 

Need to come up with a way to have a single place where futures are resolved to avoid creating the closures

bdraco:
This should fix HomeKit devices sometimes failing to pair due to timeout if we can solve it

bdraco:
Been trying to solve this for a while and can't come up with a workable design. Opened the issue so I don't forget as it too me a while to fixture out it was the closure creation pattern that was the bottleneck as it's not obvious in the profile

bdraco:
For pairing a complex HomeKit device we need to be able to deliver 283 gatt reads and writes before the timeout

bdraco:
If we change the design to be a set of triples of futures with filters and add none checks for the callback filters so we don't have pass empty callables I think we can get it fast enough that it doesn't timeout. 

We can use tuple unpacking to do extract the names so the code is more readable

bdraco:
You would think with all the effort to make all this faster that this wouldn't be a problem but it's a pretty unique scenario where the latency kills us here

bdraco:
I hate to spend the time to optimize this because it's taking away time from looking at the hot path for a very specific case. But it's a failure case so it seems worth doing

bdraco:
No matter how this gets fixed _process_packet will have to change and that will make advertising processing slower which is a problem.  

I think we can make _process_packet a factory closure which holds all the references to self in local variables which should speed it up enough that we can do an alternate dispatch path without slowing down the rest

bdraco:
If we have a complex type and a simple type in the process packet closure factory we can guard them with a simple py true check.  That should mean we only have to pay for the tuple overhead when we have a pending message.

bdraco:
Maybe it's time to write a cython wrapper.

Will write this all out and if I feel like it's too complex cython might be the way

bdraco:
Another thing we can do to reduce the latency is send sooner before creating the closure. That's safe as long as we install the the callback before the await

bdraco:
Actually all the sends are needlessly complex now that we got rid of the awaits 

send_message_callback_response probably can avoid the remove since we should only install after it doesn't fail 

We don't have to worry about it being called before since there is no await anymore

bdraco:
The read exception callbacks are also needlessly complex now since we only have one type of handler.  They can be a list of futures instead now since they are all resolved the same way

[...]

PR #474, opened 2023-07-15
Make do_append and do_stop optional in send_message_await_response_complex

bdraco:
This avoids constructing empty lambdas

related issue #470


bdraco:
running out the batteries on my devices testing this .. 🤣

PR #477, opened 2023-07-15
Speed up formatting the cipher nonce

### LINKED ISSUE #476: encryption is bottlenecked by int to bytes conversion
format_nonce

bdraco:
fixes #476

```
 % python3 timer.py
0.045173624996095896
0.07660245802253485

```
```
import timeit
from functools import partial
from struct import Struct

PACK_NONCE = partial(Struct("<LQ").pack, 0)


def call2(n: int):
    return b"\x00\x00\x00\x00" + n.to_bytes(length=8, byteorder="little")


print(timeit.timeit("call(1)", globals={"call": PACK_NONCE}))
print(timeit.timeit("call2(1)", globals={"call2": call2}))
```

### DIFF
--- a/aioesphomeapi/_frame_helper.py
+++ b/aioesphomeapi/_frame_helper.py
@@ -4,8 +4,9 @@ import logging
 from abc import abstractmethod
 from enum import Enum
 from functools import partial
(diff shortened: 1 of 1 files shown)


bdraco:
Screenshot 2023-07-15 at 9 12 05 AM

PR #478, opened 2023-07-15
Speed up timeout implementation in send_message_await_response_complex

bdraco:
Since we are already awaiting a future we can set a timeout exeception with a call_later instead of the overhead of async_timeout

related issue #470


bdraco:
It helps, but it still not enough to get the latency down enough

bdraco:
We can do the same thing in device connect for bluetooth devices which should help when we have to retry

PR #479, opened 2023-07-15
Reduce overhead to send messages that need to wait for responses

bdraco:
see #470

### DIFF
--- a/aioesphomeapi/connection.py
+++ b/aioesphomeapi/connection.py
@@ -764,7 +764,7 @@ class APIConnection:
                 self._send_pending_ping = False
 
             handlers = message_handlers.get(msg_type)
-            if handlers is not None:
+            if handlers:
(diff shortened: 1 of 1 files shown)


aioesphomeapi/connection.py
```diff
@@ -170,7 +170,7 @@ def __init__(
         self._connect_complete = False
 
         # Message handlers currently subscribed to incoming messages
-        self._message_handlers: Dict[Any, List[Callable[[message.Message], None]]] = {}
+        self._message_handlers: Dict[Any, Set[Callable[[message.Message], None]]] = {}
```
bdraco, line 173: If the set slows down the callback, revert that and remove the suppress and switch to a bare try

bdraco: its fine

PR #484, opened 2023-07-17
Remove memoryview wrapping for cameras

bdraco:
In #456 these the camera bytes were wrapped in memoryview but since the memory is not contiguous the final string has to copy anyways which makes overhead of wrapping in a memoryview is useless and counterproductive. The real speed up was not building a new bytes string every message and the cost of wrapping the memoryview was so small that I did not notice it was ineffective


bdraco:
tested with an esp32-cam

PR #486, opened 2023-07-17
Speed up noise handshake

bdraco:
The goal is to make deep sleep + encryption more viable as the current reconnect cycle can keep the device awake too long and use more power


bdraco:
I think we can make a send multiple requests in the helper so it sends all the data in one packet

PR #487, opened 2023-07-18
Speed up encrypted handshake

bdraco:
Reduces the number of round trips by one since
we can send the hello and the noise handshake
in the same packet


bdraco:
Just realized there is a long standing bug if the hello and handshake response come from in the same packet from the esp we will miss the handshake repsonse

bdraco:
> Just realized there is a long standing bug if the hello and handshake response come from in the same packet from the esp we will miss the handshake repsonse

This probably is not fixable with the current protocol design since there is no length sent in the handshake so we don't know if we have the full packet or not.  Its all smaller than the MTU anyways so unlikely to ever be a problem but technically it could be

PR #488, opened 2023-07-18
Cleanup connection send_message from earlier refactoring

bdraco:
- remove unneeded local var copy
- guard asserts with if TYPE_CHECKING
- remove unused pylint disables


bdraco:
testing is good

PR #489, opened 2023-07-18
Reduce ChaCha20CipherReuseable overhead

bdraco:
- Adds some missing typing


bdraco:
The frame helper should be split into multiple files with one for each type

Noise
Plaintext
_base?

PR #490, opened 2023-07-18
Small speed up to bluetooth message wrapper

bdraco:
The goal is to speed up HomeKit pairing

While #470 is solved for faster system, we still see some pairing timeouts with an RPI3b. 

We are running out of places where we can squeeze performance out to avoid the failures

### DIFF
--- a/aioesphomeapi/client.py
+++ b/aioesphomeapi/client.py
@@ -492,7 +492,11 @@ class APIClient:
         address: int,
         handle: int,
         request: message.Message,
-        response_type: Type[message.Message],
+        response_type: Union[
+            Type[BluetoothGATTNotifyResponse],
+            Type[BluetoothGATTReadResponse],
+            Type[BluetoothGATTWriteResponse],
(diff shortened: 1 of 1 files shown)


aioesphomeapi/client.py
```diff
@@ -468,25 +468,44 @@ def _on_home_assistant_service_response(
...
     async def _send_bluetooth_message_await_response(
         self,
         address: int,
         handle: int,
         request: message.Message,
-        response_type: Type[message.Message],
+        response_type: Union[
+            Type[BluetoothGATTNotifyResponse],
+            Type[BluetoothGATTReadResponse],
+            Type[BluetoothGATTWriteResponse],
```
bdraco, line 498: I wish we could increase the minimum python version here as the new syntax is so much better

bdraco:
Managed to get rid of the ignores as well.

Would be nicer if we could upgrade the minimum python version as its so much more readable to use the newer syntax

PR #493, opened 2023-07-19
Avoid mypy ignore when converting BLE raw advertisement

bdraco:
The protobuf and dataclasses are designed to always match order for round trips so we can use values() instead which avoids the ignore


bdraco:
This might be too slow so it might be better to keep the ignore

bdraco:
I don't like this better. protobuf just isn't very flexible here.

PR #506, opened 2023-07-27
Improve plain_text frame helper performance

bdraco:
Avoid the early bytes conversion so we do not
create multiple immutable bytes strings to get
the message type

### DIFF
--- a/aioesphomeapi/_frame_helper/base.py
+++ b/aioesphomeapi/_frame_helper/base.py
@@ -40,7 +40,7 @@ class APIFrameHelper(asyncio.Protocol):
 
     def __init__(
         self,
-        on_pkt: Callable[[int, bytes], None],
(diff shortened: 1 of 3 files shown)


bdraco:
https://github.com/protocolbuffers/protobuf/issues/10774

bdraco:
```
bdraco@MacBook-Pro-9 ~ % python3 -m timeit 'bytes(bytearray([1,2,3,4,5,6,7,8]))'
2000000 loops, best of 5: 111 nsec per loop
bdraco@MacBook-Pro-9 ~ % python3 -m timeit 'memoryview(bytearray([1,2,3,4,5,6,7,8]))'
2000000 loops, best of 5: 124 nsec per loop
```

bdraco:
improvement is not statistically significant

PR #507, opened 2023-07-27
Reduce memory churn by avoiding reprocessing BluetoothLERawAdvertisement

bdraco:
Copying the exact same structure into the dataclass was most of the expense (more than all the packet processing).

Since we want the raw data and its exactly the same structure, we can avoid the copy which is the majority of the overhead.

If we ever change the structure, we will have to go back to making a copy


bdraco:
Works great but need to make sure the typing is ok

bdraco:
Memory churn is way down.  It gets shipped right off to bluetooth-data-tools right after this and converted into bleak format

bdraco:
bleak does something similar with pyobjc for the same reason so this should be fine with the overall design

bdraco:
will run it for a day or so but it looks solid and really makes a big difference on how much this scales

bdraco:
reduces memory usage (~22MiB) as well since the constant churn is gone and our GC level stays at a lower peak now

PR #525, opened 2023-08-16
Implement zerocopy reading in the protocol

bdraco:


bdraco:
all the packets are so small that its actually slower with all the python wrapping

PR #526, opened 2023-08-16
Reduce buffer copy when we read a whole frame at once

bdraco:


bdraco:
That added a lot of complexity for ~18% faster and no difference for noise.

Not worth it

bdraco:
The packets are small and the python overhead is high.

Its likely a waste of time to try to optimize this more unless we go cython

PR #532, opened 2023-09-03
Updated fork of PR for Text input components

EternityForest:
Building on work by mauritskorse,  https://github.com/mauritskorse/aioesphomeapi, this PR adds support for the text components in the matching forked PR https://github.com/esphome/esphome/pull/5336

mauritskorse did essentially everything already, I just fixed a few merge issues and moved the messsage IDs as mentioned in the original review on the mauritskorse PR.


bdraco (approved):

tests/test_model.py
```diff
@@ -346,6 +350,7 @@ def test_user_service_conversion():
         LockInfo,
         MediaPlayerInfo,
         AlarmControlPanelInfo,
+        TextInfo,
```
bdraco, line 353: 👍

jesserockz:
Tried to slim it down, but not adding the `unique_id` field causes the library to throw a KeyError somewhere that is swallowed somewhere else and was hard to track down.

bdraco:
We are likely stuck with it for backwards compat for at least a few versions since people will downgrade for a bit (even if that technically doesn’t need to be supported)

bdraco (approved):

aioesphomeapi/model.py
```diff
@@ -757,6 +756,28 @@ class AlarmControlPanelEntityState(EntityState):
     )
 
 
+# ==================== TEXT ====================
+class TextMode(APIIntEnum):
+    TEXT = 0
+    PASSWORD = 1
+
+
+@dataclass(frozen=True)
```
bdraco, line 765: We could use the _frozen_dataclass_decorator here so it uses less ram on newer python

aioesphomeapi/model.py
```diff
@@ -757,6 +756,28 @@ class AlarmControlPanelEntityState(EntityState):
...
+class TextInfo(EntityInfo):
+    min_length: int = 0
+    max_length: int = 255
+    pattern: str = ""
+    mode: Optional[TextMode] = converter_field(
+        default=TextMode.TEXT, converter=TextMode.convert
+    )
+
+
+@dataclass(frozen=True)
```
bdraco, line 775: We could use the _frozen_dataclass_decorator here so it uses less ram on newer python

PR #535, opened 2023-09-05
Remove legacy make_ble_raw_advertisement_processor call

bdraco:
This was unneeded complexity that was left over from earlier refactoring


bdraco:
Doesn't change the execution path, so this ends up just being code churn even if it is a bit easier to follow

bdraco:
surprising the protobuf __getattr__ implementation is really really slow

PR #536, opened 2023-09-05
DO NOT MERGE: Test compiler with protobuf 4

bdraco:


bdraco:
Doesn't help

performance bottleneck is here https://github.com/protocolbuffers/protobuf/blob/dcc1f61973c8e8d54440879cefc7d511675b976a/upb/python/message.c#L979

PR #544, opened 2023-09-08
Optimize buffering to be zero copy for single packet per read case

bdraco:
Most of the time we get a whole message with the read so we can avoid copying and merging the buffer if this is the case.

this is another attempt at #525 that does not use a deque and instead tries to keep the buffer being a bytes object until we need to actually join it

### DIFF
--- a/aioesphomeapi/_frame_helper/base.py
+++ b/aioesphomeapi/_frame_helper/base.py
@@ -70,7 +70,6 @@ class APIFrameHelper(asyncio.Protocol):
             self._buffer = data
             self._buffer_len = len(data)
             return
-        _LOGGER.warning("%s: Add partial packet to buffer", self._log_name)
         # This is the expensive case since we have to create
         # a new immutable bytes object each time, but since
         # its so rare we do not want to optimize for it
@@ -84,7 +83,6 @@ class APIFrameHelper(asyncio.Protocol):
             self._buffer = None
             self._buffer_len = 0
             return
-        _LOGGER.warning("%s: Removing partial packet from buffer", self._log_name)
         if TYPE_CHECKING:
             assert self._buffer is not None
         # This is the expensive case since we have to create
--- a/aioesphomeapi/_frame_helper/base.py
+++ b/aioesphomeapi/_frame_helper/base.py
@@ -70,6 +70,7 @@ class APIFrameHelper(asyncio.Protocol):
             self._buffer = data
             self._buffer_len = len(data)
             return
+        _LOGGER.warning("%s: Add partial packet to buffer", self._log_name)
         # This is the expensive case since we have to create
         # a new immutable bytes object each time, but since
         # its so rare we do not want to optimize for it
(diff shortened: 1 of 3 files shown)


bdraco:
We almost never hit the buffering cases, so this is overly optimized for that. We should optimize to reduce bytearray to byte conversion instead 🤦

bdraco:
After running for 5 minutes we only had 238 buffering with 12 bluetooth proxies so even if we end up creating the bytes object 238 times, we avoid ~96000 bytearray to byte conversions

aioesphomeapi/_frame_helper/base.py
```diff
@@ -64,13 +64,43 @@ def _set_ready_future_exception(self, exc: Exception) -> None:
...
+        if not self._buffer:
+            self._buffer = data
+            self._buffer_len = len(data)
+            return
+        # This is the expensive case since we have to create
+        # a new immutable bytes object each time, but since
+        # its so rare we do not want to optimize for it
+        self._buffer += data
+        self._buffer_len += len(data)
+
```
bdraco, line 78: This almost never happens

aioesphomeapi/_frame_helper/base.py
```diff
@@ -64,13 +64,43 @@ def _set_ready_future_exception(self, exc: Exception) -> None:
...
+            self._buffer_len = 0
+            return
+        if TYPE_CHECKING:
+            assert self._buffer is not None
+        # This is the expensive case since we have to create
+        # a new immutable bytes object each time, but since
+        # its so rare we do not want to optimize for it
+        end_of_frame_pos = self._pos
+        self._buffer_len -= end_of_frame_pos
+        self._buffer = self._buffer[end_of_frame_pos:]
```
bdraco, line 93: This happens when the event loop is a bit behind but its not that bad since it means we only slice once per packet

aioesphomeapi/_frame_helper/noise.py
```diff
@@ -329,12 +326,12 @@ def write_packet(self, type_: int, data: bytes) -> None:
...
             ) from err
 
-    def _handle_frame(self, frame: bytearray) -> None:
+    def _handle_frame(self, frame: bytes) -> None:
         """Handle an incoming frame."""
         if TYPE_CHECKING:
             assert self._decrypt is not None, "Handshake should be complete"
         try:
-            msg = self._decrypt(bytes(frame))
+            msg = self._decrypt(frame)
```
bdraco, line 334: This makes the most difference for noise since we no longer have to make another copy here

bdraco:
the other problem is that we still do slicing which creates another copy

bdraco:
https://github.com/esphome/aioesphomeapi/pull/545 will use a memory view.. not sure if it will work

bdraco:
It works, but because of the python function overhead it isn't any faster.

If we want to improve the performance here, its time to cython since python itself is the bottleneck

PR #545, opened 2023-09-08
test with memoryview

bdraco:


bdraco:
It works, but because of the python function overhead to create the memory view it isn't any faster

PR #551, opened 2023-09-18
Clear buffer when size reaches zero instead of slicing

bdraco:
Most of the time we reach a zero buffer. Its cheaper to resize it to 0 instead of slicing


bdraco:
The python overhead still overshadows it anyways

PR #564, opened 2023-10-11
Add optional basic cython implementation for frame_helper

bdraco:
The frame helper is where most of the overhead is for processing incoming packets now that we are using the upb implementation of protobuf.

Since we process 1000s of bluetooth packets per minute, the overhead of python shipping around and decoding the bytes can be a bit high.

This adds an optional cython extension to target the hot areas of the code.  I only bothered with receiving data as I don't know if there is a use case for high volume sending (if there is we can optimize that in another pass).

If cython is not available, the build script will not cythonize and install the pure python version.

### DIFF
--- a/aioesphomeapi/_frame_helper/base.py
+++ b/aioesphomeapi/_frame_helper/base.py
@@ -20,6 +20,8 @@ SOCKET_ERRORS = (
 WRITE_EXCEPTIONS = (RuntimeError, ConnectionResetError, OSError)
 
 _int = int
+
+
 class APIFrameHelper:
     """Helper class to handle the API frame protocol."""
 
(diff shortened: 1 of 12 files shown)


aioesphomeapi/_frame_helper/base.py
```diff
@@ -19,8 +19,10 @@
 
 WRITE_EXCEPTIONS = (RuntimeError, ConnectionResetError, OSError)
 
+_int = int
 
-class APIFrameHelper(asyncio.Protocol):
+
+class APIFrameHelper:
```
bdraco, line 25: `asyncio.Protocol` is optional but since cython only allows one base class, we can removed it.

jesserockz (approved):

bdraco:
Thanks

I don't think we need wheel builds for this outside of HA since thats the only place that really cares about Bluetooth performance. If someone asks for them, I'll set up `ciwheelbuild` in the release process though.

bdraco:
Looks like our wheel builder for HA won't try to build the binary unless there is at least one binary wheel on PyPI so ... ciwheelbuild it is

PR #567, opened 2023-10-12
Update publish workflow to create wheels

bdraco:


bdraco:
This might take a few tries to get right since I'm used to doing this with poetry.

Sorry for any noise

PR #569, opened 2023-10-12
Disable wheel builds on old pypy versions as well

bdraco:
We don't support py<3.9

Don't try to build wheels one them


bdraco:
If the release time gets annoying, we can drop all the pypy wheels as it will cut it in 1/2

PR #573, opened 2023-10-14
Improve performance of processing incoming packets

bdraco:
process_packet now has cdefs to avoid having to lookup the object types every time a packet is processed

This is nearly the only place left worth optimizing

### DIFF
--- a/aioesphomeapi/connection.pxd
+++ b/aioesphomeapi/connection.pxd
@@ -59,3 +59,10 @@ cdef class APIConnection:
     cpdef _set_connection_state(self, object state)
 
     cpdef _report_fatal_error(self, Exception err)
+
+    @cython.locals(handlers=set)
+    cpdef _add_message_callback_without_remove(self, object on_message, object msg_types)
+
+    cpdef add_message_callback(self, object on_message, object msg_types)
+
+    cpdef _remove_message_callback(self, object on_message, object msg_types)
--- a/aioesphomeapi/connection.pxd
+++ b/aioesphomeapi/connection.pxd
@@ -1,5 +1,6 @@
 import cython
 
+
 cdef dict MESSAGE_TYPE_TO_PROTO
 cdef dict PROTO_TO_MESSAGE_TYPE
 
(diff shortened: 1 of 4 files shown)


aioesphomeapi/connection.pxd
```diff
@@ -0,0 +1,58 @@
...
+cdef bint TYPE_CHECKING
+
+cdef object DISCONNECT_REQUEST_MESSAGE
+cdef object PING_REQUEST_MESSAGE
+cdef object PING_RESPONSE_MESSAGE
+
+cdef object DisconnectRequest
+cdef object PingRequest
+cdef object GetTimeRequest
+
```
bdraco, line 19: I kept this pxd simple and only added types as I didn't want to make it difficult for anyone else to maintain.

aioesphomeapi/connection.py
```diff
@@ -246,24 +252,23 @@ def _cleanup(self) -> None:
...
         if self.on_stop and self._connect_complete:
-
-            def _remove_on_stop_task(_fut: asyncio.Future[None]) -> None:
-                """Remove the stop task.
-
-                We need to do this because the asyncio does not hold
-                a strong reference to the task, so it can be garbage
-                collected unexpectedly.
-                """
-                self._on_stop_task = None
```
bdraco, line 257: I removed the closure here and bound it to the object since cython doesn't support cpdef's with closures.  We could take the cpdef out as well but it seemed better to remove the closure instead as its slower in python as well

aioesphomeapi/connection.py
```diff
@@ -725,86 +729,72 @@ def _report_fatal_error(self, err: Exception) -> None:
         self._set_connection_state(ConnectionState.CLOSED)
         self._cleanup()
 
-    def _process_packet_factory(self) -> Callable[[int, bytes], None]:
+    def _process_packet(self, msg_type_proto: _int, data: _bytes) -> None:
```
bdraco, line 732: Since cython can nearly avoid the cost of `self` it didn't make sense to make it a factory anymore

bdraco, line 732: This is where all the heat is in the profile

PR #576, opened 2023-10-14
Split connection process to enable faster reconnects

### LINKED ISSUE #575: Speed up reconnecting
If the first reconnect fails, keep the zeroconf listener running and cancel the connection attempt if we are not in phase 2 as soon as we see the device. 

To make this cancel safe we need to split the connection logic into two phases

Phase 1: actual connection
Phase 2: setup connection 

Phase 1 must be cancel safe

bdraco:
Note: this isn't a breaking change for anything using the ReconnectLogic, but I figure someone might be calling the connection logic directly so I tagged it a major.

If the first connect fails, we now keep the zeroconf listener going and will cancel any in-flight connection attempts as soon as the device announces it online to mdns. This restores the connection logic to how is used to be before we did some fixes for races which prevented reconnecting from working sometimes. This version should not suffer from the same problem, and still retain the faster reconnects because it splits up the connection logic to be safer.

To accomplish this the connection process is now split into two parts:

start_connection and finish_connection

- start_connection is intended to be able to be safely cancelled without putting to much load on the ESP device since it does not do a handshake or login and is only in charge of establishing the connection
- finish_connection does the handshake, auth, and login. The reconnect logic will not cancel the in-flight connection attempt and try connecting again when the connection has already reached this state

also: zeroconf listen was broken on esphome itself and only worked in HA. It will work on both now

Testing

- [x] esphome code
- [x] HA use case

closes #575

### DIFF
--- a/aioesphomeapi/client.py
+++ b/aioesphomeapi/client.py
@@ -354,14 +354,8 @@ class APIClient:
             )
 
     async def device_info(self) -> DeviceInfo:
+        self._check_authenticated()
         connection = self._connection
-        if not connection:
-            raise APIConnectionError(f"Not connected to {self._log_name}!")
-        if not connection or not connection.is_connected:
-            raise APIConnectionError(
-                f"Connection not ready yet for {self._log_name}; "
-                f"current state is {connection.connection_state}!"
-            )
         resp = await connection.send_message_await_response(
             DeviceInfoRequest(), DeviceInfoResponse
         )
--- a/aioesphomeapi/client.py
+++ b/aioesphomeapi/client.py
@@ -103,6 +103,7 @@ from .api_pb2 import (  # type: ignore
 from .connection import APIConnection, ConnectionParams
 from .core import (
     APIConnectionError,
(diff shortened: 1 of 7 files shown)


aioesphomeapi/client.py
```diff
@@ -352,8 +352,6 @@ def _check_authenticated(self) -> None:
                 f"Authenticated connection not ready yet for {self._log_name}; "
                 f"current state is {connection.connection_state}!"
             )
-        if not connection.is_authenticated:
-            raise APIConnectionError(f"Not authenticated for {self._log_name}!")
```
bdraco, line 355: This was useless since connected isn't set unless we are past this point

bdraco:
This isn't going to work because we have 3 levels of connection

`ReconnectLogic` -> `Client` -> `Connection`

bdraco:
refactored the client as well to be a wrapper around the connection functions

bdraco:
Production test:

- [x] 89
- [x] 806
- [x] 127.0.0.1

PR #577, opened 2023-10-15
Improve consistency of name logging

bdraco:
We have three places that log the name:
- reconnect_logic
- connection
- client

The goal it to make them all the same
since various callers pass the name
in different ways to preserve compat


bdraco:
In the future it would be nice if once it was resolved it changed the logger to `name @ resolved_ip`, but the resolver is two levels deep so not fixing that here

PR #587, opened 2023-10-17
Move varuint functions into plain_text frame_helper

bdraco:
These are only used in the plain_text frame helper.  When I split up the frame helpers I forgot to move them.

### DIFF
--- a/aioesphomeapi/_frame_helper/plain_text.py
+++ b/aioesphomeapi/_frame_helper/plain_text.py
@@ -6,7 +6,6 @@ from functools import lru_cache
 from typing import TYPE_CHECKING
 
 from ..core import ProtocolAPIError, RequiresEncryptionAPIError, SocketAPIError
-from ..util import bytes_to_varuint, varuint_to_bytes
 from .base import WRITE_EXCEPTIONS, APIFrameHelper
 
 _LOGGER = logging.getLogger(__name__)
--- a/aioesphomeapi/_frame_helper/plain_text.py
+++ b/aioesphomeapi/_frame_helper/plain_text.py
@@ -39,8 +39,7 @@ def _bytes_to_varuint(value: _bytes) -> _int | None:
     """Convert bytes to a varuint."""
     result = 0
     bitpos = 0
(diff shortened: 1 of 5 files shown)


aioesphomeapi/_frame_helper/plain_text.py
```diff
@@ -10,6 +11,47 @@
...
+        if value:
+            result.append(temp | 0x80)
+        else:
+            result.append(temp)
+
+    return bytes(result)
+
+
+_cached_varuint_to_bytes = lru_cache(maxsize=1024)(_varuint_to_bytes)
+varuint_to_bytes = _cached_varuint_to_bytes
```
bdraco, line 36: We make `varuint_to_bytes` a cython object, but if we did it `_cached_varuint_to_bytes` directly it would not get the decorator

aioesphomeapi/_frame_helper/plain_text.py
```diff
@@ -10,6 +11,47 @@
...
+        val = value[pos]
+        result |= (val & 0x7F) << bitpos
+        if (val & 0x80) == 0:
+            return result
+        bitpos += 7
+    return None
+
+
+_cached_bytes_to_varuint = lru_cache(maxsize=1024)(_bytes_to_varuint)
+bytes_to_varuint = _cached_bytes_to_varuint
```
bdraco, line 53: We make `bytes_to_varuint ` a cython object, but if we did it `_cached_bytes_to_varuint ` directly it would not get the decorator

aioesphomeapi/_frame_helper/plain_text.py
```diff
@@ -2,14 +2,54 @@
...
+def _bytes_to_varuint(value: _bytes) -> _int | None:
+    """Convert bytes to a varuint."""
+    result = 0
+    bitpos = 0
+    for val in value:
+        result |= (val & 0x7F) << bitpos
+        if (val & 0x80) == 0:
+            return result
+        bitpos += 7
+    return None
```
bdraco, line 47: Probably not reachable, but existing

PR #596, opened 2023-10-21
DO NOT MERGE: Test disabling keep alives

bdraco:
This is manual testing to see if
https://community.home-assistant.io/t/esphome-didnt-respond-to-ping-request-in-time-disconnecting/608538/17 can be replicated


bdraco:
I figured disabling the keep alives would show if we had a bug in this lib but everything seems ok so likely a network issue or the ESP itself isn't actually sending the ping request (or processing the response) 
https://github.com/home-assistant/core/issues/101174

PR #597, opened 2023-10-22
Make reconnect logic state machine switches check locks

bdraco:
- When the state machine changed state we did not always check that we held the lock. We now do to ensure we avoid future regressions

- This did not uncover any existing bugs. The goal it to make the implementation safer against future refactoring


bdraco:
73.60% coverage now... we are getting there

PR #598, opened 2023-10-22
Refactor internal message handlers to use a single dispatch path

bdraco:
We had two ways of handling incoming packets:

- dispatch via registered handlers
- hard coded list of handlers

To simplify the handling of incoming packets we now register the internal handlers which leaves a single dispatch path for processing packets


bdraco:
Tested good overnight


Surprisingly we already have coverage that hits the time handler from one of the handshake set up tests I added earlier this year

PR #611, opened 2023-11-02
Use a memoryview when processing bytes in the frame helper

bdraco:
This should avoid all the pyint conversions when splicing up the headers


bdraco:
Won't work because we need a writable buffer .. which would be more expensive to create

bdraco:
https://github.com/cython/cython/pull/3247/files

bdraco:
We still have to end up converting back to bytes so it doesn't really help

PR #612, opened 2023-11-05
Refactor Bluetooth raw advs to avoid protobuf name lookups

bdraco:
The idea here is to use ListFields so we don't have to call all the name logic in [GetAttr](https://github.com/protocolbuffers/protobuf/blob/0a2fe573a292325600d1d99619a1a711a349d314/python/message.c#L999) every time

Its also possible all the work to create the tuples from ListFields is more expensive though.. so need to profile profile profile


bdraco:
Its faster, but the listcomp and distcomps make it slower because python doesn't inline those until 3.12

bdraco:
Maybe good for another look after we switch to 3.12 in HA

list_dict_comp

PR #615, opened 2023-11-07
Allow the stop callback to be cancelled when already disconnected

bdraco:
#614 fixed the case were we would cancel the handshake in progress, but it did not account for the race where we have not switched the state to connecting yet because we have not obtained the lock. While I could only get this to happen in a test its likely it possible in the real world if the timing is exactly perfect


bdraco:
Looks like we are going to end up with nearly full coverage for reconnect logic .. just one retest and PR at a time

PR #620, opened 2023-11-08
Stop building wheels for python 3.9/3.10

bdraco:
This change is only to decrease the time it takes to do a release and save space on PyPI. The release time is > 45 minutes because of the addition of aarch64 wheels which have to be done in qemu since github doesn't have native arm runners.

When esphome 2023.11.0 is released, the containers will be python 3.11 based so we will no longer need wheels for old python versions.

This does not increase the library minimum python version and wheels can still be built locally for older python versions.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -27,7 +27,7 @@ jobs:
       - name: Build wheels
         uses: pypa/cibuildwheel@v2.16.2
         env:
-          CIBW_SKIP: cp36-* cp37-* cp38-* pp*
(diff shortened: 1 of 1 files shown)


bdraco:
Merge after we are sure there will no more 2023.10.x releases of ESPHome (probably safe to merge already since we are in beta cycle for 2023.11.x)

bdraco:
Actually we can merge this now since we won't be bumping aioesphomeapi in a 2023.10.x patch build of esphome because we can't anyways

PR #622, opened 2023-11-08
Ensure passed params are always a str

bdraco:
Consumers of the api may pass string like objects that are not actually python `str` objects.  Since `expected_name` and `noise_psk` are strongly typed to `str` in `APIConnection` to allow fast string operations, we ensure they are actual `str` objects before passing them from `APIClient` to `APIConnection`

related issue https://github.com/esphome/issues/issues/5090

### DIFF
--- a/aioesphomeapi/client.py
+++ b/aioesphomeapi/client.py
@@ -249,9 +249,7 @@ def _stringify_or_none(value: str | None) -> str | None:
     the client as an Estr, but we want to pass it
     to the API as a string or None.
     """
-    if value is None:
-        return None
-    return str(value)
+    return None if value is None else str(value)
 
 
 # pylint: disable=too-many-public-methods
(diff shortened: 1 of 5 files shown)


aioesphomeapi/client.py
```diff
@@ -242,6 +242,18 @@
 ]
 
 
+def _stringify_or_none(value: str | None) -> str | None:
+    """Convert a string to a string or None.
```
bdraco, line 246:
```suggestion
    """Convert a string like object to a str or None.
```

tests/test_client.py
```diff
@@ -22,6 +22,9 @@
     SwitchCommandRequest,
     TextCommandRequest,
 )
+import asyncio
+from .common import get_mock_zeroconf,Estr
```
TheZoker, line 26:
```suggestion
from .common import get_mock_zeroconf, Estr
```

bdraco:
sadly this fails the testing check list when I pushed it to production.

I'm not sure whats wrong yet

bdraco:
I missed that a `noise_psk` of `''` is a special case and should be treated as `None` since HA will pass it as '' if its not set depending on how the entry was set up

bdraco:
fixed that in https://github.com/esphome/aioesphomeapi/pull/623

PR #625, opened 2023-11-09
Small speed up to the plaintext frame helper

bdraco:


bdraco:
I think this can be refactored to make a uint, and build it up instead

That should be much faster

bdraco:
Either way this is ~19% faster so lets merge this and refactor.

I think I can make the code a lot simpler

PR #628, opened 2023-11-09
Send hello and login asynchronously to speed up connecting

bdraco:
Current we send HelloRequest, wait for a response, than send ConnectRequest and wait for a response.

We can send the `HelloRequest` and `ConnectRequest`, and than wait for both responses instead as it reduces one round trip wait. 

This should speed up connecting with encryption enabled a little bit since it reported to be slow in https://github.com/home-assistant/core/issues/103517#issuecomment-1798017912

```
INFO Starting log output from largegarage.local using esphome API
INFO Successfully connected to largegarage in 0.061s
INFO Successful handshake with largegarage in 0.022s
[18:45:21][I][app:102]: ESPHome version 2023.11.0b2 compiled on Nov  8 2023, 23:13:11
```
```
INFO Starting log output from icemaker.local using esphome API
INFO Successfully connected to icemaker in 0.020s
INFO Successful handshake with icemaker in 0.045s
[18:46:15][I][app:102]: ESPHome version 2023.12.0-dev compiled on Nov  9 2023, 16:50:08
```

### DIFF
--- a/aioesphomeapi/client.py
+++ b/aioesphomeapi/client.py
@@ -441,8 +441,8 @@ class APIClient:
             return isinstance(msg, ListEntitiesDoneResponse)
 
         assert self._connection is not None
-        resp = await self._connection.send_message_await_response_complex(
-            ListEntitiesRequest(), do_append, do_stop, msg_types, timeout=60
+        resp = await self._connection.send_messages_await_response_complex(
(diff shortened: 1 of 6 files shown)


bdraco:
Looks like we have no coverage for wrong password

bdraco:
Manually test with wrong encryption key as well to make sure it doesn't cause a problem

PR #631, opened 2023-11-11
Add humidity and aux heat support to climate

kroimon:
Related PRs:
- esphome/esphome#5732
- home-assistant/core#103807


kroimon:
I chose `uint32` for humidity, because Home Assistant uses an integer for both the current and target humidity.
Unfortunately, this gives us no possibility to send invalid values like `NAN` would when using `float`.

I thought about either using a `float` anyway or adding something like `bool missing_current_humidity` like the `SensorStateResponse` uses.
Any input on this is highly appreciated.

bdraco:
Its ok to use a float for `humidity` and convert it when it gets to HA

kroimon:
> Its ok to use a float for `humidity` and convert it when it gets to HA

I will change the humidity type in both ESPHome and the API to `float` and convert it to `int` in HA.
That aligns it with the temperatures, allows to use NaN and maybe even future-proofs if HA decides to expand the type for higher precision.

kroimon:
All PRs updated, ready for review.

bdraco (changes requested):
CI is failing, please check

kroimon:
> Warning: aioesphomeapi/client.py:1208:4: R0912: Too many branches (13/12) (too-many-branches)

I don't see a way to restructure that, the function just follows the usual translation from `optional<T>` to protobuf messages.
If we don't want to refactor the whole class, I guess increasing the `max-branches` pylint config would be the best way to solve this warning.

bdraco:
You can do something like
`    async def bluetooth_device_connect(  # pylint: disable=too-many-locals, too-many-branches`

bdraco (approved):
LGTM. May need to change with feedback from the `esphome` PR

jesserockz (approved):
Looks good to me also

PR #632, opened 2023-11-11
Ensure zeroconf instance is closed when log runner ends

bdraco:
The log runner was using the sync close function instead of the async one


bdraco:
missing test coverage is for passed in zeroconf, which is ok considering was had 0 coverage here 10 minutes ago

PR #641, opened 2023-11-16
Refactor frame helper to avoid py conversions when processing packets

bdraco:
Since frame helper know about the `APIConnection` object now it can stay in native code the whole time to call the `process_packet` and `report_fatal_error` functions.  Its a bit more tight coupling but frame helper is already tightly coupled to connection and its all internal anyways.

This makes processing incoming packets ~21% faster

### DIFF
--- a/aioesphomeapi/_frame_helper/base.pxd
+++ b/aioesphomeapi/_frame_helper/base.pxd
@@ -1,30 +1,31 @@
 
 import cython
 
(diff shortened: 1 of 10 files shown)


bdraco:
missing coverage is existing (just a function rename)

bdraco:
We are nearly at 84% coverage now 😁

PR #642, opened 2023-11-16
Fix handling incoming data from protractor event loop

bdraco:
protractor sends us `bytearray` where we expect `bytes`

related issue https://github.com/esphome/issues/issues/5117

### DIFF
--- a/aioesphomeapi/_frame_helper/noise.py
+++ b/aioesphomeapi/_frame_helper/noise.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import binascii
 import logging
-from enum import Enum
 from functools import partial
 from struct import Struct
 from typing import TYPE_CHECKING, Any, Callable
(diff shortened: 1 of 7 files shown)


aioesphomeapi/_frame_helper/noise.py
```diff
@@ -168,7 +160,14 @@ def data_received(self, data: bytes) -> None:
...
             try:
-                self._dispatch(self, frame)
+                if self._state == NOISE_STATE_READY:
+                    self._handle_frame(frame)
+                elif self._state == NOISE_STATE_HELLO:
+                    self._handle_hello(frame)
+                elif self._state == NOISE_STATE_HANDSHAKE:
+                    self._handle_handshake(frame)
+                else:
+                    self._handle_closed(frame)
```
bdraco, line 170: I wrote this out so we it does the bytes check here to make sure we didn't have another problem deeper in the stack

PR #646, opened 2023-11-20
Ensure expected_disconnect is True when sending DisconnectResponse fails

bdraco:
If the ESP sends `DisconnectRequest` and shuts down the connection immediately before we can reply with `DisconnectResponse`, `expected_disconnect` could fail to be set because it was set after `DisconnectResponse` was sent.  I've never seen this actually happen but given the related issues below we should fix this race to eliminate it as a future problem.

related issues: https://github.com/esphome/issues/issues/4794 and https://github.com/esphome/issues/issues/5013

Although I don't think this will fix that problem since nobody reported that there was an error in the log about not being able to write the `DisconnectResponse`, but we should still fix this race.

### DIFF
--- a/aioesphomeapi/connection.py
+++ b/aioesphomeapi/connection.py
@@ -883,8 +883,11 @@ class APIConnection:
         self, _msg: DisconnectRequest
     ) -> None:
         """Handle a DisconnectRequest."""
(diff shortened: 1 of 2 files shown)


bdraco:
I kind hate to do a release for something as simple as this but better to eliminate it as a potential issue than have users having to do testing to see if it is

PR #649, opened 2023-11-21
Cleanup disconnecting handler to use asyncio.timeout

bdraco:
Previously we used asyncio.wait to avoid cancelling the task, but since we now cancel it anyways we can use asyncio.timeout


bdraco:
I looks like we still need this to be able to propagate the error upwards to the original awaiter

PR #660, opened 2023-11-22
Refactor connection checks to return APIConnection to avoid many asserts

bdraco:


bdraco:
Missing lines are existing

PR #661, opened 2023-11-22
Rename internal client connection get function to _get_connection

bdraco:
This is an internal change only to improve readability


bdraco:
The client only has two connection states, connected and not connected. It was confusing to call it authenticated as thats an internal property of the connection object

bdraco:
missing coverage is existing (its getting there)

PR #666, opened 2023-11-23
Fix race running disconnect callback in reconnect logic

bdraco:
We did not hold the lock while we ran the disconnect callback. Also it would not run if we were stopped and we disconnect which means that callers might not tear down things that were expected.

This likely was not an actual problem in production but we could have started to reconnect before the disconnect hook had finished


bdraco:
well of course there is no coverage for the disconnect callback.. 😉

PR #697, opened 2023-11-24
Fix signedness of server name check

bdraco:


bdraco:
I'll need to do another release for this since it could never be -1 

🤦 🤦 🤦 🤦 🤦 🤦 🤦 🤦 🤦 🤦 🤦

PR #698, opened 2023-11-25
Fix race in cleaning up connection

bdraco:
Because the on_stop happens in a task it has to wait one event loop iteration to remove the connection from the client

### DIFF
--- a/aioesphomeapi/client.py
+++ b/aioesphomeapi/client.py
@@ -274,13 +274,12 @@ class APIClient:
     ) -> None:
         # Hook into on_stop handler to clear connection when stopped
         self._connection = None
-        if on_stop is None:
-            return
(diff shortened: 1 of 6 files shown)


bdraco:
There is still a race in the ReconnectLogic because it runs in a task

I guess it really doesn't need to do that since nothing downstream actually needs to `await`, but its a breaking change since it would now take a callable..

need to think on that

bdraco:
While there is still a chance ha is cleaning up from the disconnect while we are reconnecting here, because we do the reconnect in two tasks ha should always be done before we finish so it's probably not a problem in practice. I think I'll avoid the breaking change and just go with what I have now as it does fix the connection being unset to late which actually can be a problem

PR #700, opened 2023-11-25
Avoid expensive debug logging checks in packet processor

### LINKED ISSUE #699: Debug logging checks happen on every packet
We should instead check if debug is enabled when creating the connection

We need a .set_debug(default true) for ha to be able to change it at runtime though

bdraco:
This should only be a breaking change for HA (we can avoid actually breaking anything by flipping the flag when HA changes the logging)

fixes #699

Testing
- [x] HA (needs a two line change, and than only a breaking change in the respect that to adjusting debug logging at run time after the integration is set up)
- [x] Esphome (not a breaking change)
- [x] logging cli (not a breaking change)

### DIFF
--- a/aioesphomeapi/client.py
+++ b/aioesphomeapi/client.py
@@ -205,6 +205,7 @@ class APIClient:
             Can be used to prevent accidentally connecting to a different device if
             IP passed as address but DHCP reassigned IP.
         """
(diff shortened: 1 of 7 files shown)


bdraco:
The two missing lines are in an existing uncovered block and are only debug logging guards

PR #707, opened 2023-11-25
Remove unreachable code in connection cleanup

bdraco:
We only set the result of _read_exception_futures in one place so there is no need to check if they are done anymore


bdraco:
These are actually reachable

PR #719, opened 2023-11-25
Speed up protobuf message type lookup

bdraco:


bdraco:
speed up is within the margin of error

not worth the code churn

closing

PR #721, opened 2023-11-26
DO NOT MERGE: Add a cache to the deserialize

bdraco:
This is a test


bdraco:
This probably isn't viable because the hit rate is likely to be too low to be worth it

bdraco:
We need it to be at least 25% to be worth it, and I expect its somewhat ~ 10%

bdraco:
2023-11-25 18:37:31.361 CRITICAL (SyncWorker_9) [homeassistant.components.profiler] Cache stats for lru_cache <cyfunction _cached_merge_from_string at 0x7fe51daf0c70> at unknown: CacheInfo(hits=430, misses=10856, maxsize=1024, currsize=1024)


cache is useless

PR #726, opened 2023-11-26
Fix reconnect logic cancelling the connection while handshaking

bdraco:
There was an unconditional cancel of the connect task in the zeroconf path


aioesphomeapi/reconnect_logic.py
```diff
@@ -226,11 +226,11 @@ async def _handle_connection_failure(self, err: Exception) -> None:
 
     def _schedule_connect(self, delay: float) -> None:
         """Schedule a connect attempt."""
-        self._cancel_connect("Scheduling new connect attempt")
```
bdraco, line 229: This was unconditionally cancelling the connection attempt in the middle of a handshake

PR #744, opened 2023-11-26
Add coverage for subscribe_voice_assistant

bdraco:


aioesphomeapi/client.py
```diff
@@ -1240,7 +1239,8 @@ def api_version(self) -> APIVersion | None:
     async def subscribe_voice_assistant(
         self,
         handle_start: Callable[
-            [str, int, VoiceAssistantAudioSettings], Coroutine[Any, Any, int | None]
+            [str, int, VoiceAssistantAudioSettingsModel],
```
bdraco, line 1242: This was typed to the protobuf model instead of the dataclass

PR #753, opened 2023-11-26
Fix voice assistant start task getting prematurely garbage collected

bdraco:


bdraco:
The task didn't actually get cancelled because it needed `nonlocal` to hold it

PR #762, opened 2023-11-28
Fix not backing off when connection requires encryption

bdraco:
related issue https://github.com/home-assistant/core/issues/104624

### DIFF
--- a/aioesphomeapi/connection.py
+++ b/aioesphomeapi/connection.py
@@ -940,15 +940,3 @@ class APIConnection:
                 )
 
         self._cleanup()
-
-    def _get_message_handlers(
-        self,
-    ) -> dict[Any, set[Callable[[message.Message], None]]]:
-        """Get the message handlers.
-
-        This function is only used for testing for leaks.
-
-        It has to be bound to the real instance to work since
-        _message_handlers is not a public attribute.
-        """
-        return self._message_handlers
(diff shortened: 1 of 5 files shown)


aioesphomeapi/connection.py
```diff
@@ -120,8 +120,8 @@ class ConnectionState(enum.Enum):
     # The handshake has been completed, messages can be exchanged
     HANDSHAKE_COMPLETE = 2
     # The connection has been established, authenticated data can be exchanged
-    CONNECTED = 2
-    CLOSED = 3
+    CONNECTED = 3
+    CLOSED = 4
```
bdraco, line 124: oops! 🤦 🤦 🤦 🤦 🤦 🤦 🤦

tests/test_reconnect_logic.py
```diff
@@ -719,6 +720,8 @@ async def on_connect() -> None:
         protocol = cli._connection._frame_helper
         send_plaintext_hello(protocol)
         send_plaintext_connect_response(protocol, False)
+        await asyncio.sleep(0)
+        await asyncio.sleep(0)
```
bdraco, line 724: The connection runs in a task so it wasn't actually finished yet but because of enum bug it looked like it was

aioesphomeapi/connection.py
bdraco: Its been that way for a while 😢

PR #763, opened 2023-11-28
Reduce duplicate code between connection and frame helper

bdraco:
- The perform_handshake code was effectively duplicated in connection since we had try functions to convert exceptions from earlier designs.
- Reduce some duplicate code between modules


aioesphomeapi/_frame_helper/base.py
```diff
@@ -65,8 +66,9 @@ def set_log_name(self, log_name: str) -> None:
         self._log_name = log_name
 
     def _set_ready_future_exception(self, exc: Exception | type[Exception]) -> None:
-        if not self._ready_future.done():
-            self._ready_future.set_exception(exc)
+        """Set the ready future to an exception."""
```
bdraco, line 69:
```suggestion
```

aioesphomeapi/connection.py
```diff
@@ -131,6 +132,46 @@ class ConnectionState(enum.Enum):
...
+
+
+handle_timeout = _handle_timeout
+
+
+def _handle_complex_message(
+    fut: asyncio.Future[None],
+    responses: list[message.Message],
+    do_append: Callable[[message.Message], bool] | None,
+    do_stop: Callable[[message.Message], bool] | None,
```
bdraco, line 161: It was a bit nicer when they were bound to the object, but there is no self use so they technically should be separate methods.

PR #783, opened 2023-11-29
Partially revert "Add humidity and aux heat support to climate (#631)"

kroimon:
Remove deprecated aux heat functionality.
I think we can safely remove these fields from the API because they were never in any production version.

See https://github.com/home-assistant/core/pull/103807#discussion_r1408794846 and https://github.com/home-assistant/architecture/discussions/932

- Related Home Assistant PR:
 home-assistant/core#104694


bdraco (approved):

aioesphomeapi/api.proto
```diff
@@ -846,7 +846,6 @@ message ListEntitiesClimateResponse {
   bool supports_target_humidity = 23;
   float visual_min_humidity = 24;
   float visual_max_humidity = 25;
-  bool supports_aux_heat = 26;
```
bdraco, line 849: Let's call this well planned that it ended up being the highest value

kroimon: Hehe, sure, all planned 😎

PR #789, opened 2023-12-10
Add happy eyeballs support (RFC 8305)

bdraco:
Adds Happy Eyeballs support (https://datatracker.ietf.org/doc/html/rfc8305)

--

This PR is a first step as I need to make sure this library works for the other projects as well. Its likely to sit for a bit while I test the library with all the projects that need it so there aren't any need to do breaking changes before all the other projects start consuming it:
- https://github.com/aio-libs/aiohttp/pull/7954
- https://github.com/Jc2k/aiohomekit/pull/349

--

replaces and closes #233

### DIFF
--- a/aioesphomeapi/client.py
+++ b/aioesphomeapi/client.py
@@ -280,7 +280,7 @@ class APIClient:
         """Set the log name of the device."""
         resolved_address: str | None = None
         if self._connection and self._connection.resolved_addr_info:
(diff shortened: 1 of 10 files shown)


bdraco:
I've just finished the code for all 3 of these PRs.

I'm doing testing aiohomekit and aioesphomeapi but I don't want to merge until I'm sure there won't be any breaking changes in the lib

PR #796, opened 2023-12-12
Add support for passing multiple addresses to the client

bdraco:
If we have multiple IP addresses for the ESPHome device, and we do not know which one we should connect to, they should be passed as `addresses` when creating the `APIClient`

Tweaks Happy Eyeballs delay to 0.1s since we expect its local

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,7 +86,7 @@ jobs:
       - run: mypy aioesphomeapi
         name: Check typing with mypy
         if: ${{ matrix.python-version == '3.11' && matrix.extension == 'skip_cython' }}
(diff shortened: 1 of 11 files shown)


.github/workflows/ci.yml
```diff
@@ -86,7 +86,7 @@ jobs:
       - run: mypy aioesphomeapi
         name: Check typing with mypy
         if: ${{ matrix.python-version == '3.11' && matrix.extension == 'skip_cython' }}
-      - run: pytest -vv --cov=aioesphomeapi --cov-report=xml --tb=native tests
+      - run: pytest -vvvs --cov=aioesphomeapi --cov-report=xml --tb=native tests
```
bdraco, line 89:
```suggestion
      - run: pytest -vv --cov=aioesphomeapi --cov-report=xml --tb=native tests
```

PR #826, opened 2024-02-16
Avoid creating tasks for starting/finishing the connection

bdraco:
We end up creating a lot of tasks when the device is in deep sleep since its trying to reconnect.  We can replace them with `async_interrupt` and futures. This simplifies the connection process and reduces the chance for races.

This also speeds up connections because we don't end up with 3 tasks to make a connection anymore, only 1

### DIFF
--- a/aioesphomeapi/connection.py
+++ b/aioesphomeapi/connection.py
@@ -627,7 +627,7 @@ class APIConnection:
         if isinstance(ex, APIConnectionError):
             return ex
         cause: BaseException | None = None
-        if isinstance(ex, CancelledError):
+        if isinstance(ex, (ConnectionInterruptedError, CancelledError)):
             err_str = f"{action.title()} connection cancelled"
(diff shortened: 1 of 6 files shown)


aioesphomeapi/connection.py
```diff
@@ -475,19 +467,14 @@ async def _connect_hello_login(self, login: bool) -> None:
...
-                tuple(messages),
-                None,
-                lambda resp: type(resp)  # pylint: disable=unidiomatic-typecheck
-                is msg_types[-1],
-                tuple(msg_types),
-                CONNECT_REQUEST_TIMEOUT,
-            )
-        except TimeoutAPIError as err:
-            self.report_fatal_error(err)
-            raise TimeoutAPIError("Hello timed out") from err
```
bdraco, line 489: This was a useless try-except-raise that actually produced a worse error message

PR #827, opened 2024-02-18
Convert functions that no longer await to be normal functions

bdraco:
These functions no longer need to be coroutines since there is nothing to be awaited. The protocol does not ACK any of these requests (the connection will drop if there is a failure) so there is no need to make them awaitables

### DIFF
--- a/aioesphomeapi/client.py
+++ b/aioesphomeapi/client.py
@@ -410,7 +410,7 @@ class APIClient:
                 entities.append(cls.from_pb(msg))
         return entities, services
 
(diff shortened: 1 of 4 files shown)


bdraco:
Checked with Jesse on this one to make sure its not going to cause a headache.

esphome shouldn't be affected by this change, only HA and bleak connector for ESPHome as ESPHome already uses the helper `esphome/components/api/client.py:from aioesphomeapi.log_runner import async_run`

bdraco:
Finished the HA PR and bleak connector PR. All good

PR #853, opened 2024-03-30
Support for Event entity messages

nohat:


jesserockz:
@nohat can you put up a draft ha core pr so we have something to reference and test with?

aioesphomeapi/model.py
```diff
@@ -262,6 +262,17 @@ def is_closed(self, api_version: APIVersion) -> bool:
         return self.position == 0.0
 
 
+# ==================== EVENT ==================
+@_frozen_dataclass_decorator
+class EventInfo(EntityInfo):
+    event_types: list[str] = converter_field(default_factory=list, converter=list)
```
jesserockz, line 268: `device_class` is missing from here

jesserockz (changes requested):
codecov seems to be broken right now in CI, but please add some pytests to ensure 100% coverage on this project is kept.

jesserockz (approved):

PR #854, opened 2024-04-04
Send/Receive Voice Assistant Audio Messages

jesserockz:
Along with new a new message to handle audio data, the `voice_assistant_version` on the DeviceInfo has been deprecated in favour of `voice_assistant_feature_flags` to let the client know the features that this device supports.

There is a `voice_assistant_feature_flags_compat(ApiVersion)` function that will return the flags, or convert the legacy version into flags for older devices.

Credit to @synesthesiam for starting this work.

Related PRs:
- https://github.com/esphome/esphome/pull/6471
- https://github.com/home-assistant/core/pull/114800


synesthesiam (approved):
Looks good to me! Probably best to break out the `end` flag rather than relying on empty bytes.

jesserockz:
> Looks good to me! Probably best to break out the `end` flag rather than relying on empty bytes.

@synesthesiam What do you mean? There is an `end` flag already in the Audio message...

synesthesiam:
I mean what you did (adding the flag) is better than what I had done 😄

PR #860, opened 2024-04-13
Remove zeroconf listener removal workaround

bdraco:
[0.132.1 fixes the problem with removing a listener from inside a listener](https://github.com/python-zeroconf/python-zeroconf/pull/1370)

HA has already bumped zeroconf. [Should wait to merge for esphome to bump it as well](https://github.com/esphome/esphome/pull/6548)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 aiohappyeyeballs>=2.3.0
 async-interrupt>=1.1.1
 protobuf>=3.19.0
-zeroconf>=0.132.1,<1.0
+zeroconf>=0.132.2,<1.0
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -1,7 +1,7 @@
 aiohappyeyeballs>=2.3.0
 async-interrupt>=1.1.1
 protobuf>=3.19.0
-zeroconf>=0.128.4,<1.0
+zeroconf>=0.132.1,<1.0
```
bdraco, line 4:
```suggestion
zeroconf>=0.132.2,<1.0
```

jesserockz (approved):

PR #870, opened 2024-05-06
Cache bytes to protobuf

bdraco:


bdraco:
I think I had tried this before but it didn't help so much because its all a bit random

2024-05-05 16:25:45.769 CRITICAL (SyncWorker_11) [homeassistant.components.profiler] Cache stats for lru_cache <cyfunction _message_from_data at 0x7f90a98846c0> at unknown: CacheInfo(hits=736, misses=10774, maxsize=1024, currsize=1024)

PR #871, opened 2024-05-07
Adding announcement flag to media player command

gnumpi:
HA marks media player calls with a specific flag if the audio to be played is an announcement.
This PR includes this flag into the media player command message.
Additionally, a new media player state "ANNOUNCING" is introduced.

related PR:
https://github.com/esphome/esphome/pull/6691
https://github.com/home-assistant/core/pull/116993


aioesphomeapi/api.proto
```diff
@@ -1125,6 +1125,7 @@ enum MediaPlayerState {
   MEDIA_PLAYER_STATE_IDLE = 1;
   MEDIA_PLAYER_STATE_PLAYING = 2;
   MEDIA_PLAYER_STATE_PAUSED = 3;
+  MEDIA_PLAYER_STATE_ANNOUNCING = 4;
```
jesserockz, line 1128: As per my comment in the esphome pr. Please also bump the api client version in this library so that the device knows the client can accept this new state.

jesserockz (changes requested):

aioesphomeapi/api.proto
gnumpi: I thought the the version is updated automatically by the CI, that's why I didn't change any yet. Can you give me a hint where I have to bump the api client version?

jesserockz: https://github.com/esphome/aioesphomeapi/blob/e97a716d44ac8759dcdf497c693d66963806bd27/aioesphomeapi/connection.py#L148

jesserockz: You can just bump the minor version to 11

jesserockz: The API version and the library version are two different things =)

gnumpi: ahh great, thanks for clarifying!

jesserockz (approved):

PR #900, opened 2024-07-04
Add support for water_heater entities.

kpfleming:


bdraco:
Is there an ESPHome PR for this ?

kpfleming:
There is not yet, although I could begin working on one. I'm not aware of any method for an ESPHome PR to build against this one, so the CI jobs on the ESPHome PR will fail (but I can build locally).

bdraco:
> There is not yet, although I could begin working on one. I'm not aware of any method for an ESPHome PR to build against this one, so the CI jobs on the ESPHome PR will fail (but I can build locally).

You should be able to build it without updating `aioesphomeapi` there as there is a separate `api.proto` in that repo

PR #921, opened 2024-08-11
Add support for creating eager tasks to reduce connect latency

bdraco:
We create tasks within tasks to connect which adds a bit of latency. We can avoid this with eager tasks. We still end up with the happyeyeballs overhead but that can be improved once cpython fixes some of the bugs in `asyncio.staggered` (See https://github.com/aio-libs/aiohttp/issues/8599)


bdraco:
need to bump mypy first https://github.com/esphome/aioesphomeapi/pull/913

PR #925, opened 2024-08-15
Add MediaPlayerSupportedFormat

synesthesiam:
Extends the media player info with a list of supported audio formats. This is used in Home Assistant to swap out media URLs going to the ESPHome device with proxy URLs that contain the converted audio.

Depends on the following PRs: 
* https://github.com/home-assistant/core/pull/123254
* https://github.com/esphome/esphome/pull/7318


synesthesiam:
@jesserockz Did I do something wrong to cause this test to fail?

jesserockz:
@synesthesiam I think you generated the files with a different proto version. Ill try fix

aioesphomeapi/model.py
```diff
@@ -820,10 +820,42 @@ class MediaPlayerCommand(APIIntEnum):
...
+        converter=MediaPlayerFormatPurpose.convert,
+    )
+
+    @classmethod
+    def convert_list(cls, value: list[Any]) -> list[MediaPlayerSupportedFormat]:
+        ret = []
+        for x in value:
+            if isinstance(x, dict):
+                if "type_" in x and "type" not in x:
+                    x = {**x, "type": x["type_"]}
```
jesserockz, line 844: I don't think this is needed. It's only used on another message where there actually is a `type` field
```suggestion
```

jesserockz (changes requested):

synesthesiam:
> @synesthesiam I think you generated the files with a different proto version. Ill try fix

Ok, if not I'll try to upgrade and regenerate the files tomorrow.

synesthesiam:
@jesserockz I have libprotoc version 3.21.12. Is this too old? Too new?

jesserockz:
You can use docker to generate too
```bash
docker run -v "$PWD":/aioesphomeapi ghcr.io/esphome/aioesphomeapi-proto-builder:latest
```

jesserockz (approved):

synesthesiam:
Do I need to add a version bump too?

PR #932, opened 2024-08-29
Try to set TCP_QUICKACK where support for that option is available

nkinnan:
Looking at network traces (on windows), the ESP responds with more data within 20ms to every TCP ACK of an API packet, but the python API consumer doesn't set the proper socket option to disable delayed ack and so windows delays for 40-50ms before sending an ack back.

Delayed ack waits for more data so multiple packets can be acked at the same time, which in most circumstances would be more efficient, except the esp has limited buffer space and must retain the tcp send buffer in case a retry is required if the packet doesn't get acked at all, so by disabling delayed ack on the other side of the API connection, packets can be acked immediately and this loop can be reduced to 1/2 to 1/3rd of it's current latency allowing 2-3x the send buffers to be dispatched over the TCP connection

This will result in less dropped states/events over the API 

In testing on a HAOS virtualbox, I found that TCP_QUICKACK is already set system-wide and so this change is redundant there, but it is helpful for anywhere else the service might run which doesn't have that same default.  It is also useful for the esphome commandline tool which uses the API to get logs after an OTA push. 

Turns out this setting isn't supported in python on windows, only on linux, so I went ahead and fixed that too: https://github.com/python/cpython/pull/123478

### DIFF
--- a/aioesphomeapi/connection.py
+++ b/aioesphomeapi/connection.py
@@ -372,6 +372,13 @@ class APIConnection:
         self._socket = sock
         sock.setblocking(False)
         sock.setsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)
+        try:
+            sock.setsockopt(socket.IPPROTO_TCP, socket.TCP_QUICKACK, 1)
+        except AttributeError:
+            _LOGGER.debug(
+                "%s: TCP_QUICKACK not supported",
+                self.log_name,
+            )
         self._increase_recv_buffer_size()
         self.connected_address = sock.getpeername()[0]
 


bdraco:
This probably makes sense now that we group the writes and don't have cases where we send a lot of writes anymore where waiting for them all to be acked would make send since most of the time the messages we are sending are small and they get sent together.

bdraco:
We group on reconnect https://github.com/esphome/aioesphomeapi/blob/87cb7c3a9b49ec17805c6bdcafca66186ca18cad/aioesphomeapi/connection.py#L465

bdraco:
These might be a bit slower with this change though since they aren't grouped and we probably get 3 acks instead of 1 here
https://github.com/home-assistant/core/blob/ff9937f942828c18c82e8e6ff66ca9f0004fcea8/homeassistant/components/esphome/manager.py#L536

nkinnan:
@bdraco This fix addresses a behavior much further down the network stack at the TCP protocol layer, it doesn't have anything to do with messages being sent over that connection.  If you're on the discord, DM me or start a thread and @ me (nkinnan) and I can explain better.

nkinnan:
@bdraco It will *always* be faster to ack immediately than to wait, because of how TCP is implemented on the ESP.  While the ESP has sent a packet (or packets, comprising a single send buffer) out, it can not service any more events or messages into the send queue until it gets an ack back on the packet(s) it already sent.  The faster that ACK gets back, the faster it can release the buffer and start processing more data to send.

nkinnan:
@bdraco The primary limiting factor in ESP throughput over the API is ack delay.  Throughput in bytes/sec that the ESP can send out is (tcp_send_buffer_size_in_bytes * number_of_acks) since the buffer is locked until an ack comes back.  And number_of_acks is limited by how fast the other end can send them back.  By default systems will wait 50-200ms to send an ack, limiting the ESP to only being able to send 5-20 tcp_send_buffer's worth of data.  Fortunately HAOS has TCP_QUICKACK set system-wide so it's already responding as fast as possible, which is lucky for us.  But other OSes don't, and will be forced to communicate more slowly.

bdraco:
I'm worried about additional Wi-Fi traffic with Bluetooth proxies. In this case the ESP is sending a group of raw advertisements every few ms to Home Assistant.  If it can send multiple with only having to get back one ACK that this change will generate more traffic. If the ESP going to have wait for the ACK on every send than this change should not impact the amount of traffic.

nkinnan:
Currently, no matter how full or empty the tcp send buffer is, once send() is called, that buffer is locked until an ack comes back, and no further data can be sent.

bdraco:
I'm not sure `TCP_QUICKACK` is going to be defined on all platforms we support. Will need to check that.

nkinnan:
@bdraco I tested that, it's why there is a try/catch around it.  In fact I linked a PR to cpython where I added support for it to the windows python runtime.  But the try/catch handles the situation now before that eventually gets released.

bdraco:
I'm going to test this on my mac laptop as well

bdraco (approved):
Tested with MacOS 👍 
Confirmed its already turned on with HAOS 👍 

Thanks @nkinnan

nkinnan:
Thanks for getting it in so quickly :)

PR #939, opened 2024-08-29
Adds voice assistant announce

synesthesiam:
Adds `VoiceAssistantAnnounce` message and `send_voice_assistant_announce`

* https://github.com/esphome/esphome/pull/7377
* https://github.com/home-assistant/core/pull/125157
* https://github.com/esphome/voice-kit/pull/64


synesthesiam:
Leaving as a draft since this is still under discussion.

bdraco:
I fixed codecov so it should report correctly now

aioesphomeapi/client.py
```diff
@@ -1286,6 +1289,13 @@ def subscribe_voice_assistant(
             ]
             | None
         ) = None,
+        handle_announce_finished: (
```
jesserockz, line 1292: Wouldn't it be better to have this callback in the `send_voice_assistant_announce` method?

synesthesiam: That should work fine, and would make the HA side simpler.

synesthesiam: Moved the wait inside aioesphomeapi

aioesphomeapi/model.py
```diff
@@ -1291,6 +1292,11 @@ class VoiceAssistantAudioData(APIModelBase):
     end: bool = False
 
 
+@_frozen_dataclass_decorator
+class VoiceAssistantAnnounceFinished(APIModelBase):
+    success: bool
```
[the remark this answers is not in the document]
jesserockz: We always define the default value here. protobuf removes the values from the message if it is the default. For example if you sent a success=false message back, its only gonna send the message id to save on bytes.

jesserockz (approved):

PR #948, opened 2024-09-03
Remove id from ``MediaPlayerSupportedFormat``

jesserockz:
This `Message` is not sent or received by itself, but always as part of the `ListEntitiesMediaPlayerResponse`, so it doesn't need an id

- https://github.com/esphome/esphome/pull/7406


bdraco (approved):
Thanks @jesserockz

PR #949, opened 2024-09-04
Avoid slice when the reader wants the full buffer

bdraco:


bdraco:
we can never avoid the slice

PR #977, opened 2024-10-13
Add codspeed benchmarks for Bluetooth messages

bdraco:
Start benchmarking with codspeed to ensure we don't regress performance


bdraco:
It looks like we might be accessing the protobuf object more than expected

PR #978, opened 2024-10-13
DNM: Test benchmarks with cython disabled

bdraco:


bdraco:
Looks like its working as expected 👍

PR #992, opened 2024-11-06
Bump protobuf to 25.5 (aka PyPi 4.25.5)

bdraco:
Home Assistant currently uses protobuf==5.28.3 and only one major version ahead is technically supported so we seeing issues like https://github.com/home-assistant/core/issues/129052 https://github.com/home-assistant/core/issues/130169

Note that the problem only appears if C extensions are not working and protobuf falls back to the pure python version


hcjehg:
It seems this has failed on 
@github-actions **CI / py 3.12 on ubuntu-latest (skip_cython) (pull_request) Failing after 1m**

:-)

hcjehg:
It seems this still fails :-)
CI / py 3.12 on ubuntu-latest (skip_cython) (pull_request) Failing after 1m

PR #1040, opened 2025-01-29
Avoid buffer copy on read

bdraco:


bdraco:
no way to avoid the copy because

PR #1053, opened 2025-02-02
Add more sending benchmarks

bdraco:
https://github.com/protocolbuffers/protobuf/blob/243b023c4530c1349437f036feb96cd053c0a8b2/python/convert.c#L192 seems to be called when setting bool and it takes an much longer than expected time. Might be a bug in protobuf


bdraco:
It looks like they already fixed it https://github.com/protocolbuffers/protobuf/commit/9e44913d434426d4e3dbccbc8ad26ea2c4edd382

but we have an old version?

PR #1072, opened 2025-02-16
Resolve hosts in parallel

bdraco:
If the system doesn't support mDNS, or mDNS will take more time than the timeout to respond, we would never try DNS.  Instead we now resolve mDNS and DNS in parallel.

Additionally we had a 30s timeout to resolve, but the mDNS timeout was only 3s so we would end up timing out too quickly.

Sadly this is a lot more complex than I would like because we support resolving any number of host names at the same time so we can potentially have `X hosts` * `X IPs` results.

### DIFF
--- a/aioesphomeapi/connection.py
+++ b/aioesphomeapi/connection.py
@@ -52,13 +52,8 @@ from .core import (
     UnhandledAPIConnectionError,
 )
 from .model import APIVersion, message_types_to_names
+from .util import asyncio_timeout
 from .zeroconf import ZeroconfManager
-
-if sys.version_info[:2] < (3, 11):
-    from async_timeout import timeout as asyncio_timeout
-else:
-    from asyncio import timeout as asyncio_timeout
-
 
 _LOGGER = logging.getLogger(__name__)
 
--- a/aioesphomeapi/connection.py
+++ b/aioesphomeapi/connection.py
@@ -589,8 +589,12 @@ class APIConnection:
 
     async def _do_connect(self) -> None:
         """Do the actual connect process."""
+        addrs = self._params.addresses
+        port = self._params.port
+        zc_manager = self._params.zeroconf_manager
         await self._connect_socket_connect(
-            hr.async_addrinfos_from_ips(self._params.addresses, self._params.port)
+            hr.async_addrinfos_from_ips(addrs, port)
+            or hr.async_addrinfos_from_zeroconf_cache(zc_manager, addrs, port)
             or await self._connect_resolve_host()
         )
 
--- a/aioesphomeapi/connection.py
+++ b/aioesphomeapi/connection.py
@@ -46,7 +46,6 @@ from .core import (
     PingFailedAPIError,
     ProtocolAPIError,
     ReadFailedAPIError,
-    ResolveAPIError,
     SocketAPIError,
     SocketClosedAPIError,
     TimeoutAPIError,
@@ -95,7 +94,6 @@ DISCONNECT_CONNECT_TIMEOUT = 5.0
 
 DISCONNECT_RESPONSE_TIMEOUT = 10.0
 HANDSHAKE_TIMEOUT = 30.0
-RESOLVE_TIMEOUT = 30.0
 CONNECT_REQUEST_TIMEOUT = 30.0
 
 # The connect timeout should be the maximum time we expect the esp to take
@@ -312,20 +310,6 @@ class APIConnection:
     def set_debug(self, enable: bool) -> None:
         """Enable or disable debug logging."""
         self._debug_enabled = enable
-
-    async def _connect_resolve_host(self) -> list[hr.AddrInfo]:
-        """Step 1 in connect process: resolve the address."""
-        try:
-            async with asyncio_timeout(RESOLVE_TIMEOUT):
-                return await hr.async_resolve_host(
-                    self._params.addresses,
-                    self._params.port,
-                    self._params.zeroconf_manager,
-                )
-        except asyncio_TimeoutError as err:
-            raise ResolveAPIError(
-                f"Timeout while resolving IP address for {self.log_name}"
-            ) from err
 
     async def _connect_socket_connect(self, addrs: list[hr.AddrInfo]) -> None:
         """Step 2 in connect process: connect the socket."""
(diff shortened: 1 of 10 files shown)


aioesphomeapi/host_resolver.py
```diff
@@ -159,6 +160,28 @@ def async_addrinfos_from_ips(ips: list[str], port: int) -> list[AddrInfo] | None
...
+    addrs: list[AddrInfo] = []
+    for host in hosts:
+        if (
+            not host_is_local_name(host)
+            or not (short_host := host.partition(".")[0])
+            or not (service_info := _make_service_info_for_short_host(short_host))
+            or not service_info.load_from_cache(aiozc.zeroconf)
+        ):
+            # If any host is not in the cache, return None
+            # so we can take teh slow path
```
bdraco, line 179:
```suggestion
            # so we can take the slow path
```

aioesphomeapi/host_resolver.py
```diff
@@ -186,44 +210,92 @@ def _async_ip_address_to_addrinfo(ip: IPv4Address | IPv6Address, port: int) -> A
...
+            continue
 
-    if not addrs:
-        if zc_error:
-            # Only show ZC error if getaddrinfo also didn't work
-            raise zc_error
-        raise ResolveAPIError(
-            f"Could not resolve host {hosts} - got no results from OS"
-        )
+        coros: Coroutine[Any, Any, list[AddrInfo]] = []
```
bdraco, line 251:
```suggestion
        coros: list[Coroutine[Any, Any, list[AddrInfo]]] = []
```

aioesphomeapi/host_resolver.py
```diff
@@ -66,16 +67,16 @@ async def _async_zeroconf_get_service_info(
             f"Cannot start mDNS sockets: {exc}, is this a docker container without "
             "host network mode?"
```
bdraco, line 68: We need to move the creation of the instance to the resolve step so we don't make multiple since multiple names are supported

bdraco:
Do we really need to support resolving multiple names at once. This makes this so much more complex

bdraco:
If not, we should bump the major version and get rid of the multi resolving of different hosts at the same time

bdraco:
Back to research before I can continue

bdraco:
Looks like its impossible to know if more than one host is ever used elsewhere so will continue on this path

aioesphomeapi/host_resolver.py
```diff
@@ -186,44 +194,127 @@ def _async_ip_address_to_addrinfo(ip: IPv4Address | IPv6Address, port: int) -> A
...
+
+        # We got a result for a host, cancel
+        # any other tasks trying to resolve
+        # it as we are done with that host
+        for host in finished_hosts:
+            for task in host_tasks.pop(host, ()):
+                resolve_task_to_host.pop(task, None)
+                task.cancel()
+                with suppress(asyncio.CancelledError):
+                    await task
```
bdraco, line 313: Need tests for when one finishes before the other for both cases

aioesphomeapi/host_resolver.py
```diff
@@ -186,44 +194,127 @@ def _async_ip_address_to_addrinfo(ip: IPv4Address | IPv6Address, port: int) -> A
...
+    """Resolve hosts in parallel."""
+    resolve_task_to_host: dict[asyncio.Task[list[AddrInfo]], str] = {}
+    host_tasks: defaultdict[str, set[asyncio.Task[list[AddrInfo]]]] = defaultdict(set)
+    exceptions: list[BaseException] = []
+    aiozc: AsyncZeroconf | None = None
+
+    if manager:
+        try:
+            aiozc = manager.get_async_zeroconf()
+        except Exception as original_exc:
```
bdraco, line 262: OSError is enough

aioesphomeapi/host_resolver.py
```diff
@@ -159,6 +145,28 @@ def async_addrinfos_from_ips(ips: list[str], port: int) -> list[AddrInfo] | None
     return None
 
 
+def async_addrinfos_from_zeroconf_cache(
+    zeroconf_manager: ZeroconfManager | None, hosts: list[str], port: int
+) -> list[AddrInfo] | None:
+    """Convert a list of IPs to AddrInfos."""
+    if not zeroconf_manager or not zeroconf_manager.has_instance:
+        return None
+    aiozc = zeroconf_manager.get_async_zeroconf()
```
bdraco, line 154: needs to handle OSError and return `None`

https://github.com/esphome/aioesphomeapi/pull/1072/files#diff-4b9b26ad12930f8b24b706ca0fc82f78bd64ccdbdaf24936c3b6e780781289c4R260

aioesphomeapi/connection.py
```diff
@@ -589,8 +589,12 @@ def _async_pong_not_received(self) -> None:
...
     async def _do_connect(self) -> None:
         """Do the actual connect process."""
+        addrs = self._params.addresses
+        port = self._params.port
+        zc_manager = self._params.zeroconf_manager
         await self._connect_socket_connect(
-            hr.async_addrinfos_from_ips(self._params.addresses, self._params.port)
+            hr.async_addrinfos_from_ips(addrs, port)
+            or hr.async_addrinfos_from_zeroconf_cache(zc_manager, addrs, port)
             or await self._connect_resolve_host()
```
bdraco, line 598: Move these into the resolver and let the resolver take a timeout

aioesphomeapi/host_resolver.py
```diff
@@ -186,44 +172,160 @@ def _async_ip_address_to_addrinfo(ip: IPv4Address | IPv6Address, port: int) -> A
...
+                )
+            )
+
+        coros.append(_async_resolve_host_getaddrinfo(host, port))
+
+        for coro in coros:
+            task = create_eager_task(coro)
+            if task.done():
+                if exc := task.exception():
+                    exceptions.append(exc)
```
bdraco, line 301: Looks like we don't have a test for when we get an exception right away

aioesphomeapi/host_resolver.py
```diff
@@ -186,44 +172,160 @@ def _async_ip_address_to_addrinfo(ip: IPv4Address | IPv6Address, port: int) -> A
...
+                new_exc.__cause__ = original_exc
+                exceptions.append(new_exc)
+
+        if aiozc:
+            short_host = host.partition(".")[0]
+            service_info = _make_service_info_for_short_host(short_host)
+            if service_info.load_from_cache(aiozc.zeroconf) and (
+                addr_infos := service_info_to_addr_info(service_info, port)
+            ):
+                resolve_results[host].extend(addr_infos)
```
bdraco, line 247: Need a test for fetching from cache

aioesphomeapi/host_resolver.py
```diff
@@ -186,44 +172,160 @@ def _async_ip_address_to_addrinfo(ip: IPv4Address | IPv6Address, port: int) -> A
...
+    if len(resolve_results) != len(hosts):
+        # If we have not resolved all hosts yet, we need to do some network calls
+        try:
+            try:
+                async with asyncio_timeout(timeout):
+                    await _async_resolve_host(
+                        hosts, port, resolve_results, exceptions, aiozc, timeout
+                    )
+            except asyncio.TimeoutError as err:
+                raise ResolveTimeoutAPIError(
```
bdraco, line 258: Need a test for when we timeout

bdraco:
@coderabbitai review

PR #1100, opened 2025-03-08
Avoid protobuf lookup to find MergeFromString

bdraco:
The `PyUpb_Message_GetAttr` looks a bit heavy because it looks in the message first and falls back to finding `MergeFromString`. It make sense to do it only once per message type. Its a bit unexpected that protobuf implemented this as a lookup for a known method.

### DIFF
--- a/aioesphomeapi/connection.pxd
+++ b/aioesphomeapi/connection.pxd
@@ -119,7 +119,7 @@ cdef class APIConnection:
 
     cdef void send_messages(self, tuple messages)
 
-    @cython.locals(handlers=set, handlers_copy=set)
+    @cython.locals(handlers=set, handlers_copy=set, klass_merge=tuple)
     cpdef void process_packet(self, unsigned int msg_type_proto, object data)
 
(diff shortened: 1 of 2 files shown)


aioesphomeapi/connection.pxd
```diff
@@ -6,6 +6,8 @@ from ._frame_helper.base cimport APIFrameHelper
 cdef dict MESSAGE_TYPE_TO_PROTO
 cdef dict PROTO_TO_MESSAGE_TYPE
 
+cdef object MERGE_FROM_STRING
+
```
bdraco, line 10:
```suggestion
```

aioesphomeapi/connection.py
```diff
@@ -55,6 +55,8 @@
 from .util import asyncio_timeout
 from .zeroconf import ZeroconfManager
 
+MERGE_FROM_STRING = message.Message.MergeFromString
+
```
bdraco, line 59:
```suggestion
```

PR #1105, opened 2025-03-11
Improve buffer read performance

bdraco:
`__Pyx_PyBytes_AsString` -> `__Pyx_PyBytes_FromStringAndSize` is significantly faster than `PySequence_GetSlice` and since we already have a bounds check with `if self._buffer_len < new_pos:` we don't need to worry about reading past the buffer.

Note that we have no benchmark for noise protocol so we need to add some before merging this PR to validate the improvement

The speaker buffer looks to be about 16KiB from `esphome/components/voice_assistant/voice_assistant.cpp` so it should help that as well

Screenshot 2025-03-10 at 7 59 33 PM

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,13 +13,13 @@ repos:
     - id: pyupgrade
       args: [--py39-plus]
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.9.9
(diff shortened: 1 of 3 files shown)


bdraco:
github seems to be broken now as the jobs are stuck

https://github.com/esphome/aioesphomeapi/actions/runs/13782166824?pr=1105

PR #1110, opened 2025-03-12
Reduce noise buffer copy overhead

bdraco:
Screenshot 2025-03-11 at 9 38 21 PM

Its about ~4% better of the decrypt path (note the benchmark includes the test overhead and encrypt overhead as well)

### DIFF
--- a/aioesphomeapi/_frame_helper/noise.py
+++ b/aioesphomeapi/_frame_helper/noise.py
@@ -183,9 +183,7 @@ class APINoiseFrameHelper(APIFrameHelper):
                     )
                 )
                 return
-            msg_size_high = header[1]
-            msg_size_low = header[2]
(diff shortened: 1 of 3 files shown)


aioesphomeapi/_frame_helper/noise.py
```diff
@@ -357,14 +358,23 @@ def _handle_frame(self, frame: bytes) -> None:
...
             return
+        msg_length = len(msg)
+        msg_cstr = msg
+        if msg_length < 4:
+            self._handle_error_and_close(
+                ProtocolAPIError(
+                    f"{self._log_name}: Decrypted message too short: {msg_length} bytes"
+                )
+            )
+            return
```
bdraco, line 369: We don't actually have coverage for this case but it could happen in theory.... although its very unlikely

PR #1139, opened 2025-04-06
Reduce buffer copy in the frame helper

bdraco:


bdraco:
Its only 1% faster.  The trade-off is its easier to make memory safety bugs so it seems not worth it

PR #1143, opened 2025-04-08
Reduce buffer extraction overhead

bdraco:
Safer version of #1139 which does not store the
unsafe C derivative of temporary Python reference
in the object


bdraco:
.. and it makes no difference. All the overhead is memcpy_avx_unaligned_erms

PR #1145, opened 2025-04-08
Reattempt memoryview once again

bdraco:


bdraco:
The problem with small packets is creating the memoryview object is more expensive than the memcpy

PR #1153, opened 2025-04-15
Split packet generation code into its own module

bdraco:
Usually I would split it into two modules here but each one has a load cost so I put it all in one


aioesphomeapi/_frame_helper/packets.pxd
```diff
@@ -0,0 +1,26 @@
+from .noise_encryption cimport EncryptCipher
+
+cdef object varuint_to_bytes
+
+cdef _varuint_to_bytes(int value)
```
bdraco, line 5:
```suggestion
cpdef _varuint_to_bytes(int value)
```

PR #1155, opened 2025-04-16
Move cythonized client into client_base

bdraco:
- Renames client_callbacks to client_base and includes the base part of the client there. This allows us to add cythonize types for the connection


aioesphomeapi/client_base.pxd
```diff
@@ -0,0 +1,34 @@
+import cython
+
+# from .connection cimport APIConnection, ConnectionParams
```
bdraco, line 3: FIXME: circular cython import

PR #1157, opened 2025-04-16
Ensure set_cached_name_if_unset accepts subclassed str

bdraco:
The Cython type was too strict here : https://github.com/esphome/esphome/pull/8576#issuecomment-2808000287

### DIFF
--- a/aioesphomeapi/client_base.py
+++ b/aioesphomeapi/client_base.py
@@ -299,7 +299,7 @@ class APIClientBase:
 
     def _set_name_from_device(self, name: str_) -> None:
         """Set the name from a DeviceInfo message."""
-        self.cached_name = str(name)  # Make be Estr from esphome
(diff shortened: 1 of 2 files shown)


aioesphomeapi/client_base.py
```diff
@@ -297,16 +297,15 @@ def _set_log_name(self) -> None:
         if self._connection is not None:
             self._connection.set_log_name(self.log_name)
 
-    def _set_name_from_device(self, name: str) -> None:
+    def _set_name_from_device(self, name: str_) -> None:
         """Set the name from a DeviceInfo message."""
-        self.cached_name = name
+        self.cached_name = str(name)  # Make be Estr from esphome
```
bdraco, line 302:
```suggestion
        self.cached_name = str(name)  # May be Estr from esphome
```

PR #1158, opened 2025-04-16
Avoid decoding protobuf messages when there are no subscribers

bdraco:
If we are going to throw away the message, don't bother decoding it


bdraco:
it turned out we almost never get messages we don't have a handler for

PR #1161, opened 2025-04-20
Rewrite packet generation in Cython

bdraco:


bdraco:
It's only the Python call bytes converter.  That could probably be rewritten in cython and the rest doesn't matter because modern asyncio has zero copy writes

PR #1162, opened 2025-04-20
Write packet header builder in Cython

bdraco:


bdraco:
Interesting that protobuf is so slow that this doesn't really matter than much that the python code is inefficent

bdraco:
5% isn't worth the maint burden of maintaining the separate Cython code

PR #1174, opened 2025-05-05
Regenerate profile data with cryptography 45.0.1

bdraco:
This will merge as an empty change since we don't pin `cryptography` to ensure the next PR doesn't get unexpected performance data attributed to it.

ESPHome https://github.com/esphome/esphome/pull/8826 and Home Assistant https://github.com/home-assistant/core/pull/145121 have already been bumped so we don't need to require a new version as most users will get it.

Benchmarks were initially slower than expected because openssl is old and we get `LazyEvpCipherAead` instead of `EvpCipherAead`

Test run with https://github.com/pyca/cryptography/pull/12857


.github/workflows/ci.yml
```diff
@@ -133,7 +133,7 @@ jobs:
 
   benchmarks:
     name: Run benchmarks
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
```
bdraco, line 136:
```suggestion
    runs-on: ubuntu-22.04
```

.github/workflows/ci.yml
```diff
@@ -146,6 +146,7 @@ jobs:
         shell: bash
         run: |
             pip3 install -r requirements/base.txt -r requirements/test.txt
+            pip3 install https://koston.org/cryptography-45.0.0.dev1-cp311-abi3-manylinux_2_34_x86_64.whl
```
bdraco, line 149:
```suggestion
```

requirements/base.txt
```diff
@@ -3,6 +3,5 @@ async-interrupt>=1.2.0
 protobuf>=4
 zeroconf>=0.143.0,<1.0
 chacha20poly1305-reuseable>=0.13.2
```
bdraco, line 5:
```suggestion
chacha20poly1305-reuseable>=0.13.2
cryptography>=43.0.0
```

requirements/base.txt
```diff
@@ -3,6 +3,6 @@ async-interrupt>=1.2.0
...
-cryptography>=43.0.0
+cryptography>=45.0.1
```
bdraco, line 6:
```suggestion
cryptography>=43.0.0
```

PR #1181, opened 2025-05-06
Fast nonce implementation for encryption

bdraco:
Keep the nonce calculation in native code to avoid all the python overhead.


bdraco:
pointer math was nicer but won't work on big endian

PR #1146, opened 2025-04-10
[API] Sub devices

dala318:
# What does this implement/fix?

Adds the possibility to group entities on one physical device into virtual devices (in Home Assistant), handy when one ESP serves several purposes.

Read more in related ESP Home PR

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Code quality improvements to existing code or addition of tests
- [ ] Other

**Related issue or feature (if applicable):**

- fixes https://github.com/esphome/feature-requests/issues/1335

**Pull request in [esphome](https://github.com/esphome/esphome) (if applicable):**

- esphome/esphome#8544

## Checklist:
  - [ ] The code change is tested and works locally.
  - [x] If api.proto was modified, a linked pull request has been made to [esphome](https://github.com/esphome/esphome) with the same changes.
  - [ ] Tests have been added to verify that the new code works (under `tests/` folder).

PR #1229, opened 2025-06-23
Fix DeviceInfo round-trip serialization after areas/devices fields additions

bdraco:

This PR fixes an issue with `DeviceInfo` round-trip serialization that was introduced when the `areas`, `devices`, and `area` fields were added in #1146. 

When using `DeviceInfo.to_dict()` followed by creating a new `DeviceInfo` instance from that dictionary (a common pattern for mocking in tests), the code would fail with an `AttributeError` because the `area` field converter expected a protobuf object but received a dictionary.

The fix adds a `convert` method to `AreaInfo` that properly handles both dictionary and protobuf inputs, similar to how the `convert_list` methods already work for the list fields.

## Types of changes

- [ ] Other

- fixes round-trip serialization issue introduced in #1146

- N/A - This only affects the Python model classes, not the protobuf definitions

## Checklist:

### DIFF
--- a/aioesphomeapi/model.py
+++ b/aioesphomeapi/model.py
@@ -153,8 +153,7 @@ class AreaInfo(APIModelBase):
     def convert(cls, value: Any) -> AreaInfo:
         if isinstance(value, dict):
             return cls.from_dict(value)
-        else:
-            return cls.from_pb(value)
+        return cls.from_pb(value)
 
 
 @_frozen_dataclass_decorator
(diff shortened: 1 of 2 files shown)


aioesphomeapi/model.py
```diff
@@ -149,6 +149,13 @@ def convert_list(cls, value: list[Any]) -> list[AreaInfo]:
                 ret.append(AreaInfo.from_pb(x))
         return ret
 
+    @classmethod
+    def convert(cls, value: Any) -> AreaInfo:
+        if isinstance(value, dict):
+            return cls.from_dict(value)
+        else:
+            return cls.from_pb(value)
```
bdraco:
```suggestion
        return cls.from_pb(value)
```

PR #1239, opened 2025-07-05
Optimize connection handshake for modern ESPHome devices using encryp…

bdraco:

This PR optimizes the connection setup process by not waiting for the hello response when connecting without a password. Instead, we register a handler for the HelloResponse message similar to how we handle other internal messages like PingRequest and DisconnectRequest.

This change reduces connection latency for devices that don't require authentication by allowing the connection process to continue immediately after sending the hello message, rather than blocking until the response arrives.

**Note:** ESPHome deprecated password authentication in February 2023 in favor of encryption, making this optimization increasingly beneficial as more devices use encryption-only authentication.

## Types of changes

- [ ] Other

N/A

N/A - No api.proto changes required

## Checklist:


tests/test_client.py
```diff
@@ -1248,6 +1248,14 @@ async def test_device_info(
 ) -> None:
     """Test fetching device info."""
     client, connection, transport, protocol = api_client
+
+    # With the callback mechanism, the client name is updated quickly after hello response
+    # Wait a moment for the hello response to be processed
+    await asyncio.sleep(0.1)
```
bdraco:
```suggestion
    await asyncio.sleep(0)
```

tests/test_client.py
```diff
@@ -1979,6 +1987,13 @@ async def test_set_debug(
     caplog.set_level(logging.DEBUG)
 
     client.set_debug(True)
+
+    # With the callback mechanism, the client name is updated quickly after hello response
+    await asyncio.sleep(0.1)
```
bdraco:
```suggestion
    await asyncio.sleep(0)
```

bdraco:
we can't skip waiting for the initial response because we need to detect if the device requires encryption by checking if it sends a noise protocol hello instead of a plain text hello response.

PR #1268, opened 2025-07-21
Make CameraImageResponse and SubscribeLogsResponse zero_copy

bdraco:

Currently we have manually written methods for these messages in esphome. We want to remove the manual code and use these objects directly since we generate code for them. We had manual methods to avoid the copy but since we can mark them zero copy and generate different code, we avoid the duplicate code

## Types of changes

- [ ] Other

- fixes <link to issue>

## Checklist:
  - [ ] The code change is tested and works locally.


bdraco:
We may not need this as I think we can make ALL of them zero copy!

PR #264, opened 2022-09-29
Change has_bluetooth_proxy to a version

jesserockz:

PR #1376, opened 2025-10-02
[climate] Implement feature flags field to replace multiple bools

kbx81:

Implements a feature flags system for climate entities to replace multiple individual boolean capability flags. This addresses https://github.com/home-assistant/core/issues/153331 by introducing a new `REQUIRES_TWO_POINT_TARGET_TEMPERATURE` flag to distinguish between devices that optionally support dual setpoint vs. those that require it.

## Changes

- Adds `feature_flags` field (uint32) to `ListEntitiesClimateResponse` in api.proto
- Introduces `ClimateFeature` IntFlag enum with flags for all climate capabilities
- Deprecates 5 boolean fields (but keeps them for backward compatibility):
  - `supports_current_temperature`
  - `supports_two_point_target_temperature`
  - `supports_action`
  - `supports_current_humidity`
  - `supports_target_humidity`
- Adds new `REQUIRES_TWO_POINT_TARGET_TEMPERATURE` flag to handle mandatory dual setpoint
- Bumps API version from 1.12 to 1.13

## Breaking Changes

**IMPORTANT**: This PR is marked as breaking to alert API consumers that migration is required, even though nothing breaks immediately.

The boolean fields (`supports_current_temperature`, `supports_two_point_target_temperature`, `supports_action`, `supports_current_humidity`, `supports_target_humidity`) are **deprecated and will be removed in a future release** (likely with another major version bump).

### Why this is marked breaking:
1. **To fix the bug** (https://github.com/home-assistant/core/issues/153331), consumers need access to the new `REQUIRES_TWO_POINT_TARGET_TEMPERATURE` flag, which is only available via the feature flags system
2. **To enable future migration**: This gives API consumers advance notice to migrate their code now, before the deprecated fields are actually removed
3. **Future compatibility**: When the deprecated fields are removed in a future release, code that has already migrated will not be affected

### What will break later when fields are removed:
- Any code still directly accessing the deprecated boolean fields

**All code must migrate to using the feature flags system now** to access the new functionality and avoid breakage in the future.

**Migration required:**

```python
# Old code (incomplete with API 1.13+, missing new flags):
if climate_info.supports_current_temperature:
    process_current_temp()

# New code (works with all API versions and includes all flags):
flags = climate_info.supported_feature_flags_compat(api_version)
if flags & ClimateFeature.SUPPORTS_CURRENT_TEMPERATURE:
    process_current_temp()

# Access new functionality only available via feature flags:
if flags & ClimateFeature.REQUIRES_TWO_POINT_TARGET_TEMPERATURE:
    handle_required_dual_setpoint()
```

The `supported_feature_flags_compat(api_version)` method handles both old and new API versions:
- For API < 1.13: Maps the old boolean fields to feature flags
- For API >= 1.13: Returns the `feature_flags` field directly (includes all flags, old and new)

See also:
- https://github.com/esphome/esphome/pull/10987
- https://github.com/home-assistant/core/pull/153507

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

- fixes <link to issue>

## Checklist:


aioesphomeapi/api.proto
```diff
@@ -996,6 +996,7 @@ message ListEntitiesClimateResponse {
   float visual_min_humidity = 24;
   float visual_max_humidity = 25;
   uint32 device_id = 26 [(field_ifdef) = "USE_DEVICES"];
+  bool requires_two_point_target_temperature = 27;
```
kbx81: https://github.com/esphome/esphome/pull/10987#discussion_r2400165202

bdraco: We usually end up turning bools into flags later so might be cleaner to start out that way like #74 #264 #1285

protobuf varint encoding takes up the same or almost the same space for a bool and a feature flag of 1 and we can always add to it later

kbx81: I like that approach a lot better. I see in #264 Jesse straight-up changed a `bool` to a `uint32` -- would that be safe to do here, as well? Or better to just add a "flags" field and a comment to indicate that `supports_two_point_target_temperature` has been deprecated?

bdraco: This is brand new though? so we can just change it

kbx81: Haha, sorry -- yes, the one here is new...but there is also the previously-existing flag on line 975, which this new flag complements. I was asking if they can be safely consolidated into a uint32 on line 975 (as Jesse had swapped a `bool` to a `uint32` in #264) without adding a new field here. 🙂

bdraco: I don't think it can be safety done as there may be other consumers of it.  We didn't have that concern with Bluetooth Proxy as there is no practical use outside of HA. I'd add a new flag and we can deprecate the old one at some point.

bdraco (approved):

aioesphomeapi/model.py
```diff
@@ -643,6 +652,7 @@ class ClimatePreset(APIIntEnum):
 
 @_frozen_dataclass_decorator
 class ClimateInfo(EntityInfo):
+    feature_flags: int = 0
```
bdraco, line 655: We can add a compat call like we have for other infos similar to supported_presets_compat and others

bdraco: https://github.com/kbx81/aioesphomeapi/blob/73d44ccdd9d110da56f8ef16faf149dbd433c376/aioesphomeapi/model.py#L234

kbx81: Oh! Great catch! Yes, I'll do that -- it'll simplify the HA side. 😄

kbx81: Ok, I think I got this in [`faec940` (#1376)](https://github.com/esphome/aioesphomeapi/pull/1376/commits/faec940a389091ccc76a29e3e86ad5a3b65c2a58)

PR #1436, opened 2025-11-27
Add device defined action responses

jesserockz:

Add client-side support for user-defined service responses. This allows ESPHome devices to send response data back to the client when executing user-defined services via the `api.respond` action.

**Changes:**
- Added `call_id` and `return_response` fields to `ExecuteServiceRequest`
- Added new `ExecuteServiceResponse` message (id=131) with `call_id`, `success`, `error_message`, and `response_data` fields
- Added `SupportsResponseType` enum (NONE, OPTIONAL, ONLY, STATUS) to indicate service response capability
- Added `supports_response` field to `ListEntitiesServicesResponse` / `UserService` model
- Updated `execute_service()` to accept optional `on_response` callback and `return_response` parameters
- Callback auto-unsubscribes after receiving the matching response

**Usage:**
```python
def handle_response(response: ExecuteServiceResponse):
    print(f"Success: {response.success}, Data: {response.response_data}")

client.execute_service(
    service,
    data={"arg": value},
    on_response=handle_response,
    return_response=True,
)
```

Related PRs:
- esphome/esphome#12136
- home-assistant/core#157393
- esphome/esphome-docs#5688

## Types of changes

- [ ] Other

- esphome/esphome#XXXX

## Checklist:

### DIFF
--- a/aioesphomeapi/client_base.pxd
+++ b/aioesphomeapi/client_base.pxd
@@ -26,6 +26,7 @@ cdef str _stringify_or_none(object value)
 cdef class APIClientBase:
 
     cdef public set _background_tasks
+    cdef public object _call_id_counter
     cdef public APIConnection _connection
     cdef public bint _debug_enabled
     cdef public object _loop


aioesphomeapi/client.py
```diff
@@ -1312,10 +1314,21 @@
...
+        return_response: bool | None = None,
     ) -> None:
         connection = self._get_connection()
-        req = ExecuteServiceRequest(key=service.key)
+        # If return_response is not explicitly set, derive from on_response
+        if return_response is None:
+            return_response = on_response is not None and call_id != 0
+        req = ExecuteServiceRequest(
+            key=service.key, call_id=call_id, return_response=return_response
+        )
```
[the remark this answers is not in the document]
bdraco: If it never arrives its a bug in the device, and if the connection is torn down we get a new one, and at which point all the message handlers get cleared

aioesphomeapi/client.py
```diff
@@ -1312,10 +1314,21 @@ def update_command(
         )
 
     def execute_service(
-        self, service: UserService, data: ExecuteServiceDataType
+        self,
+        service: UserService,
+        data: ExecuteServiceDataType,
+        *,
+        call_id: int = 0,
```
bdraco: Might be a bit safer to make this a sentinel value so its harder to forget/mis-use the api

bdraco: I'm not 100% on the flow here, but does the caller need to know about the `call_id`?

```python
      if on_response is not None:
          self._call_id_counter += 1
          call_id = self._call_id_counter
          return_response = True
      else:
          call_id = 0
          return_response = False
```

bdraco: Hmm.. I guess you could have multiple HA instances controlling the device.

bdraco: Either way you have collisions. Probably still a cleaner API to have aioesphomeapi have the counter logic instead of HA having to know about it

aioesphomeapi/client.py
```diff
@@ -166,6 +167,7 @@
 from .util import create_eager_task
 
 _LOGGER = logging.getLogger(__name__)
+_CALL_ID_COUNTER = itertools.count(1)
```
bdraco: Shouldn't this be per device ?

aioesphomeapi/client_base.py
```diff
@@ -286,6 +288,7 @@ def __init__(
         self.cached_name: str | None = None
         self._background_tasks: set[asyncio.Task[Any]] = set()
         self._loop = asyncio.get_running_loop()
+        self._call_id_counter = itertools.count(1)
```
bdraco, line 291: should be added to client_base.pxd as well as object type

PR #1438, opened 2025-11-27
Fix numeric-only hostnames being misinterpreted as IP addresses

swoboda1337:
## Summary

When a hostname like `1234.local` has its `.local` suffix stripped for the getaddrinfo fallback, the resulting `"1234"` string gets interpreted by `socket.getaddrinfo()` as a decimal IP address (`0.0.4.210` in this case).

This caused OTA uploads to fail for devices with numeric-only names, as the connection would be attempted to the wrong IP address.

**Root cause demonstration:**
```python
>>> import socket
>>> socket.getaddrinfo('1234', 3232, socket.AF_INET)
[(<AddressFamily.AF_INET: 2>, <SocketKind.SOCK_STREAM: 1>, 6, '', ('0.0.4.210', 3232)), ...]
```

The decimal number 1234 equals `0*256³ + 0*256² + 4*256 + 210 = 0.0.4.210` in IPv4.

## Changes

- Skip the stripped-hostname fallback in `_async_resolve_host()` when the hostname consists only of digits
- Added two tests to verify the fix

## Test plan

- [x] Verified with `pytest tests/test_host_resolver.py` - all 33 tests pass
- [x] New test `test_resolve_host_numeric_hostname_no_stripped_fallback` verifies numeric hostnames don't trigger getaddrinfo on the stripped name
- [x] New test `test_resolve_host_alphanumeric_hostname_uses_stripped_fallback` verifies alphanumeric hostnames still use the fallback

Fixes esphome/esphome#12143

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/aioesphomeapi/host_resolver.py
+++ b/aioesphomeapi/host_resolver.py
@@ -343,10 +343,15 @@ async def _async_resolve_host(
             # is resolvable via regular DNS (violates RFC 6762 but improves compatibility)
             if host.endswith((".local", ".local.")):
                 stripped = _remove_local_suffix(host)
-                _LOGGER.debug(
-                    "Also adding DNS resolution without .local suffix for %s", stripped
-                )
-                coros.append(_async_resolve_host_getaddrinfo(stripped, port))
+                # Skip getaddrinfo for purely numeric hostnames - they get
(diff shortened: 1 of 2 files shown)


swoboda1337:
Thanks

PR #1466, opened 2025-12-23
Add version parameter to build_unique_id

bdraco:

Adds an optional `version` parameter to `build_unique_id()` to allow callers to choose between using `object_id` (version 1, default) or `name` (version 2) when constructing unique IDs.

```python
def build_unique_id(formatted_mac: str, entity_info: EntityInfo, *, version: int = 1) -> str:
```

**Version 1 (default):** `{mac}-{entity_type}-{object_id}`
**Version 2:** `{mac}-{entity_type}-{name}`

Version 2 preserves the full entity name including Unicode characters, which can be useful for distinguishing entities that would otherwise have identical `object_id` values after sanitization.

The docstring includes detailed rationale for future reference.

## Types of changes

- [ ] Other

- Related to https://github.com/esphome/backlog/issues/76

- N/A

## Checklist:

### DIFF
--- a/aioesphomeapi/model.py
+++ b/aioesphomeapi/model.py
@@ -1824,17 +1824,61 @@ _TYPE_TO_NAME = {
 }
 
 
-def build_unique_id(formatted_mac: str, entity_info: EntityInfo) -> str:
+def build_unique_id(
+    formatted_mac: str, entity_info: EntityInfo, *, version: int = 1
+) -> str:
     """Build a unique id for an entity.
 
-    This is the new format for unique ids which replaces the old format
-    that is included in the EntityInfo object. This new format is used
-    because the old format used the name in the unique id which is not
-    guaranteed to be unique. This new format is guaranteed to be unique
-    and is also more human readable.
+    Version history and rationale:
+    -----------------------------
+
+    **Version 1 (default, legacy):** Uses `object_id` which is the entity name
+    after applying `snake_case()` + `sanitize()` transformations.
+
+        Format: `{mac}-{entity_type}-{object_id}`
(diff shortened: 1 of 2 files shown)


aioesphomeapi/model.py
```diff
@@ -1824,17 +1824,61 @@ class VoiceAssistantTimerEventType(APIIntEnum):
...
+      See https://github.com/esphome/esphome-webserver/issues/160
+      and https://github.com/esphome/esphome-webserver/issues/153
+
+    The `object_id` concept is planned for removal. See
+    https://github.com/esphome/backlog/issues/76 for the full history.
+
+    **Version 2:** Uses the entity `name` directly without mangling.
+
+        Format: `{mac}-{entity_type}-{name}`
+        Example: `aabbccddeeff-sensor-Temperature Sensor`
```
arturpragacz, line 1865: Does the `name` here include something from the subdevice to make it possible to distinguish between same-named entities in different subdevices?

Also, if I'm understanding correctly, the intention is that in the new validation, character `/` (slash) will be the one to be disallowed in the names, so maybe it would be a good idea to use it here as a separator. So like:
`aabbccddeeff/sensor/Temperature Sensor`

bdraco: Technically we already have a collision issue because the device id is appended to the end as `@device_id`

https://github.com/home-assistant/core/blob/2e3eb0f9aff08cd8d6999db04aaed0b29b5a2b80/homeassistant/components/esphome/entry_data.py#L102

To avoid that we should be preventing `@` from being used in ESPHome as well, however I think we could migrate around that in the future since we know `/` is now disallowed, but something to fix on the HA side.

PR #1490, opened 2026-01-29
[api] configurable keepalive interval

rwrozelle:

Adds the attribute server_keepalive_interval to HelloResponse and use it to update the connections _keep_alive_interval, after HelloResponse.  This allows the server(device) to define the interval.

## Types of changes

- [ ] Other

- fixes <link to issue>

- esphome/esphome#13619

**Pull request in [esphome-docs](https://github.com/esphome/esphome-docs) with documentation (if applicable):**

- esphome/esphome-docs#6005

## Checklist:
  Need help on updating tests, not sure how they are done.

### DIFF
--- a/tests/test_connection.py
+++ b/tests/test_connection.py
@@ -717,7 +717,6 @@ async def test_connect_with_server_keepalive_interval(
     await connect_task
 
     assert conn.is_connected is True
-    assert conn._keep_alive_interval == 20
 
 
 async def test_force_disconnect_fails(


aioesphomeapi/api.proto
```diff
@@ -127,6 +127,7 @@ message HelloResponse {
 
   // The name of the server (App.get_name())
   string name = 4;
+  uint32 server_keepalive_interval = 5;
```
bdraco, line 130: Is there ever plans to change it once connected? If so we should get a new message. If not its fine here

rwrozelle: @bdraco I took this back to draft while I think about this and the code rabbit comment.  The edge case I was considering was a steady state of needing the ping chattiness to be at a lower threshold than the sensor status updates to lower overall request/response, I have not considered any need for transition after initialization.  My draft implementation for server side interval is based on a compiler conditional, so there also I considered it an initialization setting and not changeable.  Would growing the api to include a "PingIntervalRequest/Response" be worth it for updating the interval on the client and/or server at user discretion?  That is what I'm struggling with since api is a foundational component and growing it should not be done without thought as to value.

aioesphomeapi/connection.py
```diff
@@ -533,6 +533,10 @@ def _process_hello_resp(self, resp: HelloResponse) -> None:
             self.received_name = received_name
             self.set_log_name(self.received_name)
 
+        if keepalive := resp.server_keepalive_interval:
+            self._keep_alive_interval = keepalive
+            _LOGGER.debug("keep_alive_interval = %d", keepalive)
+
```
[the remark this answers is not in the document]
rwrozelle: @bdraco, I am unclear as to what the ratio between keepalive_interval and keepalive_timeout should be if keepalive_interval is variable.  The first is how often the client queries (ping), the second is how long a wait for the response (pong).  My draft code hardcodes this to 20*4.5 = 90, so even if there is an increase of the interval from 20 all the way to 180, the client still only waiting up to 90 seconds for a pong after the ping is sent.  I think this is reasonable because the intent of increasing the ping interval is that there is a high level of trust that the server is communicating through other request/response cycles to prove it is alive, not that there has been a corresponding increase in the latency of any given request/response cycle.  I did not include a server_keepalive_timeout parameter in the HelloResponse because of this thinking.

rwrozelle:
Close this and come back at a later point, just focus on the client side interval for now without an aioesphomeapi changes

PR #1495, opened 2026-02-12
[serial_proxy] New component

kbx81:

API modifications as required for esphome/esphome#13944

## Types of changes

- [ ] Other

- fixes <link to issue>

## Checklist:

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -28,23 +28,23 @@ jobs:
         uses: actions/checkout@v6
       -
         name: Log in to docker hub
-        uses: docker/login-action@v3.7.0
+        uses: docker/login-action@v4.0.0
         with:
           username: ${{ secrets.DOCKER_USER }}
           password: ${{ secrets.DOCKER_PASSWORD }}
       -
         name: Log in to the GitHub container registry
-        uses: docker/login-action@v3.7.0
+        uses: docker/login-action@v4.0.0
         with:
             registry: ghcr.io
             username: ${{ github.actor }}
             password: ${{ secrets.GITHUB_TOKEN }}
       -
         name: Set up QEMU
-        uses: docker/setup-qemu-action@v3.4.0
+        uses: docker/setup-qemu-action@v4
       -
         name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v3.12.0
+        uses: docker/setup-buildx-action@v4.0.0
       -
         name: Build and Push
         uses: docker/build-push-action@v6.19.2
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -41,7 +41,7 @@ jobs:
           python-version: "3.12"
       - name: Set up QEMU
         if: ${{ matrix.qemu }}
-        uses: docker/setup-qemu-action@v3
+        uses: docker/setup-qemu-action@v4
         with:
           platforms: all
           # This should be temporary
@@ -64,7 +64,7 @@ jobs:
             echo "CIBW_BUILD=${{ matrix.pyver }}*" >> $GITHUB_ENV
           fi
       - name: Build wheels
-        uses: pypa/cibuildwheel@v3.3.1
+        uses: pypa/cibuildwheel@v3.4.0
         env:
           CIBW_SKIP: cp36-* cp37-* cp38-* cp39-* cp310-* pp* ${{ matrix.musl == 'musllinux' && '*manylinux*' || '*musllinux*' }}
           CIBW_BEFORE_ALL_LINUX: apt-get install -y gcc || yum install -y gcc || apk add gcc


aioesphomeapi/api.proto
```diff
@@ -2488,3 +2497,87 @@ message InfraredRFReceiveEvent {
...
+  bool dtr = 3;                 // Current DTR pin state
+}
+
+// Flush the serial port (block until all TX data is sent)
+message SerialProxyFlushRequest {
+  option (id) = 144;
+  option (source) = SOURCE_CLIENT;
+  option (ifdef) = "USE_SERIAL_PROXY";
+
+  uint32 instance = 1;          // Instance index (0-based)
```
bdraco: Maybe make this a shared message that can do requests to the device with nothing other than the request type ie

SerialProxyRequest

request_type enum

Than if we have others in the future we don't have more protobuf message bloat

kbx81: [`8db2bdc` (this PR)](https://github.com/esphome/aioesphomeapi/pull/1495/commits/8db2bdce15c58128252f631be8def3087aeeb432)

aioesphomeapi/client_base.py
```diff
@@ -211,6 +215,20 @@ def on_infrared_rf_receive_event(
...
+    msg: SerialProxyDataReceived,
+) -> None:
+    on_data(SerialProxyDataReceivedModel.from_pb(msg))
+
+
+def on_serial_proxy_get_modem_pins_response(
+    on_modem_pins: Callable[[SerialProxyModemPins], None],
+    msg: SerialProxyGetModemPinsResponse,
+) -> None:
+    on_modem_pins(SerialProxyModemPins.from_pb(msg))
```
bdraco: Doesn't look like this is wired up?

aioesphomeapi/client.py
```diff
@@ -497,6 +509,136 @@ def infrared_rf_transmit_raw_timings(
...
+            )
+        )
+
+    def serial_proxy_flush(
+        self,
+        instance: int,
+    ) -> None:
+        """Flush the serial port (block until all TX data is sent)."""
+        self._get_connection().send_message(
+            SerialProxyRequest(
```
bdraco, line 638: We could add a `SerialProxyFlushResponse` if the deadlock scenario puddly raised could ever actually happen

aioesphomeapi/client.py
```diff
@@ -497,6 +509,136 @@ def infrared_rf_transmit_raw_timings(
...
+        self,
+        instance: int,
+        baudrate: int,
+        *,
+        flow_control: bool = False,
+        parity: SerialProxyParity = SerialProxyParity.NONE,
+        stop_bits: int = 1,
+        data_size: int = 8,
+    ) -> None:
+        """Configure UART parameters for a serial proxy instance."""
```
bdraco, line 522: Would be nice to validate stop_bits and data_size here before it gets to device

aioesphomeapi/api.proto
```diff
@@ -2517,3 +2537,94 @@ message InfraredRFReceiveEvent {
...
+
+// Response with current modem control pin states
+message SerialProxyGetModemPinsResponse {
+  option (id) = 143;
+  option (source) = SOURCE_SERVER;
+  option (ifdef) = "USE_SERIAL_PROXY";
+
+  uint32 instance = 1;          // Instance index (0-based)
+  bool rts = 2;                 // Current RTS pin state
+  bool dtr = 3;                 // Current DTR pin state
```
bdraco: Do we ever need cts/dsr ?

probably not for z-wave/zigbee but good to think about now

kbx81: Would it be more wise to make this a "flags" or maybe "line_states" field so it's easier to add more signals later on if/when we need them?

puddly: There are about 10 pins total so you could make it a bitmap or something. The various OS APIs sort of are inconsistent with how you set them: POSIX lets you set or clear a set of flags at once but has no API to atomically set the entire state in one go. Others have separate APIs for each pin.

kbx81: Indeed, there are a whole bunch of lines/signals; I'm reasonably confident saying that most just aren't used today. The two we have here now are the only two I've seen used in any reasonably modern hardware; ESP boards use them as triggers for the reset and boot pins, not actually for flow control. I gather that a lot of other microcontroller boards are doing something similar.

aioesphomeapi/client.py
kbx81: How would we handle invalid values here? A quick look suggests we don't really do validation like this directly in the client.

bdraco: raise ValueError if its outside of min <= range <= max

kbx81: [`8939402` (this PR)](https://github.com/esphome/aioesphomeapi/pull/1495/commits/8939402f185ca1902af5c188425bab71df409ba5)

aioesphomeapi/client_base.py
kbx81: Removed in [`d859238` (this PR)](https://github.com/esphome/aioesphomeapi/pull/1495/commits/d859238a1708011d60cad47c8dae40481162b00b)

aioesphomeapi/client.py
kbx81: Interested in @puddly's thoughts on the usefulness of this

puddly: Ah, I didn't see this comment was more recent.

IMO I think a flush failure should render the serial port "dead" and be propagated as a disconnect reason. The port is broken and there's nothing the client can do about it other than reconnect.

kbx81: Ah yeah, that makes sense...let me see what I can do to communicate that back up the chain

kbx81: Challenge: only IDF returns a meaningful result from `flush`; other platforms do not. This means that `serial_proxy` would only be able to report the result of `flush` when used on an ESP (other than 8266). If we build it, the status returned will be misleading (and always just "OK") on other platforms.

Not quite sure how to proceed with this -- any thoughts?

bdraco: We could return a status of ASSUME_SUCCESS or UNKNOWN on other platforms

aioesphomeapi/client.py
```diff
@@ -497,6 +509,138 @@ def infrared_rf_transmit_raw_timings(
...
+                on_data,
+            ),
+            (SerialProxyDataReceived,),
+        )
+
+    def serial_proxy_set_modem_pins(
+        self,
+        instance: int,
+        *,
+        line_states: int = 0,
```
puddly, line 568: I've seen a lot of applications just do something like this when interacting with lines:
```python
ser.dtr = True
```

So they toggle one line at a time, leaving the others intact. What do you think about an API like?

```python
high_line_states: int = 0,
low_line_states: int = 0,
```

That would expose an API to both toggle individual pins (leaving the rest alone) but also set the entire state at once.

PR #1586, opened 2026-04-15
Add temperature unit for climate and water heater

jhenkens:

Exposes temperature unit for climate and water_heater, with corresponding changes in home-assistant/core, aioesphome, esphome, and esphome-docs, to enable native Fahrenheit measurements on those entities. Usecase is to enable accurate temperature settings on a Fahrenheit hot tub, controlled via RS485-UART. Rather than round tripped from F -> C -> C, with various floating point errors and weirdnesses, this just makes it stay F the entire way and work much more reliably.

Tests still need to be added where appropriate - that's on the todo - but the code has been validated. 

When esphome is updated, but core is not, as long as you do not set the temperature unit on a climate/water heater, everything works the same as it currently is. If you do set it, you will convert from F -> "F", as home assistant will think the incoming value from ESPHome is C, without reading the UOM. If ESPHome does not have F specified, it continues to default to C.

When HA is updated, but ESPHome is on an old version, it maintains legacy behavior of treating all climate/water_heaters as using C.

## Types of changes

- [ ] Other

- fixes <link to issue>

https://github.com/home-assistant/core/pull/168261
https://github.com/esphome/aioesphomeapi/pull/1586
https://github.com/esphome/esphome/pull/15815
https://github.com/esphome/esphome-docs/pull/6463

## Checklist:


setup.py
```diff
@@ -41,7 +41,7 @@
     long_description = readme_file.read()
 
 
-VERSION = "44.17.0"
+VERSION = "0.1.0"
```
bdraco: Please remove this change

aioesphomeapi/api.proto
```diff
@@ -1025,6 +1025,13 @@ message CameraImageRequest {
   bool stream = 2;
 }
 
+// ==================== TEMPERATURE UNIT ====================
+enum TemperatureUnit {
+  TEMPERATURE_UNIT_CELSIUS = 0;
+  TEMPERATURE_UNIT_FAHRENHEIT = 1;
+  TEMPERATURE_UNIT_KELVIN = 2;
```
bdraco, line 1032: I'd leave kelvin out for now. I don't think any climate controls or water heaters measure in kelvin and at this point it's a YAGNI. It can always be added later if there is a proven case.

bdraco (changes requested):
Please see comments above

aioesphomeapi/api.proto
bdraco: Looks like water heater does support it https://developers.home-assistant.io/docs/core/entity/water-heater?_highlight=water

Actually they both use the same unit converter, but I don't see any water heaters in HA core that actually use kelvin

setup.py
bdraco:
```suggestion
VERSION = "44.17.0"
```

jhenkens: A bot is doing this.

Guessing because I used main on my fork.

jhenkens:
@bdraco Sorry for the thrash. Closing this PR in favor of https://github.com/esphome/aioesphomeapi/pull/1600 due to `main` GHA bumping version.

Only change is addressing the Copilot feedback.

aioesphomeapi/api.proto
jhenkens: Will address in https://github.com/esphome/aioesphomeapi/pull/1600 if desired - but am I hearing right that you'd prefer to keep it?

PR #1589, opened 2026-04-15
Guard log formatters against EntityInfo type mismatches

### LINKED ISSUE #1588: state_log_formatter._format_climate crashes with AttributeError when info is WaterHeaterInfo
### Summary

`_format_climate` in `state_log_formatter.py` crashes the log-runner connection with `AttributeError: 'WaterHeaterInfo' object has no attribute 'supports_two_point_target_temperature'` when `entity_info` lookups pair a `ClimateState` with a `WaterHeaterInfo`. The log viewer then reconnects and crashes again on the next state publish in an endless loop.

The device itself is unaffected; only the CLI log stream is broken.

### My setup

I'm running the `esphome-econet` external component on a Rheem heat pump water heater. The base package defines both a legacy `platform: econet` climate entity (`econet_climate`) and a newer `platform: template` water heater entity (`econet_water_heater`). **I'm intentionally using the legacy climate entity** — all my Home Assistant automations target `climate.econet_hpwh` and I never migrated off it because I didn't want to redo a bunch of automations that work fine today. The legacy entity is a supported path; `esphome-econet` keeps it specifically because users like me rely on it, and I'm not looking for them to remove it.

### Environment

- `aioesphomeapi==44.15.0` (pinned by `esphome==2026.4.0`)
- Python 3.12 (ESPHome add-on container)
- Reproducer: Rheem heat pump water heater via [`econet_heatpump_water_heater.yaml`](https://github.com/esphome-econet/esphome-econet/blob/main/econet_heatpump_water_heater.yaml)

### Traceback

```
ERROR Fatal error: protocol.data_received() call failed.
Traceback (most recent call last):
  File "/usr/local/lib/python3.12/asyncio/selector_events.py", line 1027, in _read_ready__data_received
    self._protocol.data_received(data)
  File "aioesphomeapi/_frame_helper/plain_text.py", line 48, in APIPlaintextFrameHelper.data_received
  File "aioesphomeapi/_frame_helper/plain_text.py", line 75, in APIPlaintextFrameHelper.data_received
  File "aioesphomeapi/connection.py", line 1104, in APIConnection.process_packet
  File "aioesphomeapi/client_base.py", line 76, in client_base.on_state_msg
  File "/usr/local/lib/python3.12/site-packages/aioesphomeapi/log_runner.py", line 125, in on_state
    text = format_state_log(state, info)
  File "/usr/local/lib/python3.12/site-packages/aioesphomeapi/state_log_formatter.py", line 334, in format_state_log
    return formatter(state, info)
  File "/usr/local/lib/python3.12/site-packages/aioesphomeapi/state_log_formatter.py", line 225, in _format_climate
    if info and info.supports_two_point_target_temperature:
AttributeError: 'WaterHeaterInfo' object has no attribute 'supports_two_point_target_temperature'
```

### Root cause

`log_runner.subscribe_log_states_runner` builds the info lookup keyed only by entity key:

\`\`\`python
entity_info: dict[int, EntityInfo] = {e.key: e for e in entities}
\`\`\`

Dispatch in `format_state_log` uses `type(state)` as the dispatch key and passes whatever info it finds for that key:

\`\`\`python
formatter = _STATE_FORMATTERS.get(type(state))
return formatter(state, info)
\`\`\`

This assumes `type(state)` and `type(info)` are always consistent for a given key. When they aren't, `_format_climate` (and any formatter reading entity-type-specific attributes without a type check) crashes.

In my case, both `econet_climate` and `econet_water_heater` use `name: None` in the base package and end up sharing an object_id → same 32-bit key hash. In `entity_info`, the later entry wins, so `entity_info[K]` becomes a `WaterHeaterInfo` while the device still publishes a `ClimateStateResponse` (decoded as `ClimateState`) for that same key. Dispatch routes `ClimateState → _format_climate`, the info is `WaterHeaterInfo`, and `info.supports_two_point_target_temperature` explodes.

The econet package's pattern is legitimate and I don't want them to change it. **Regardless of what any external component does, the log formatter shouldn't crash the API connection on a type mismatch** — this is a latent robustness bug that any pattern producing a state/info type mismatch on a shared key could trigger.

### Impact

- `esphome logs` / log-runner crashes every state publish (~30s cycle on this device).
- Home Assistant's normal operation is unaffected — HA doesn't use `format_state_log`, so automations and entity control on `climate.econet_hpwh` keep working.
- The ESP32 itself never reboots (verified 15+ minutes of continuous state publishes via HA history, `device_tracker` never left `home`).

### Suggested fix

Make `_format_climate` (and any other formatter that reads info-type-specific fields) defensive about the info type — either an `isinstance(info, ClimateInfo)` check returning `None` on mismatch, or `getattr(info, "supports_two_point_target_temperature", False)`. Same pattern for any other formatter reading entity-type-specific attributes. Alternatively, verify `type(state)` / `type(info)` compatibility in `log_runner.on_state` before calling `format_state_log`.

### Reproduction

1. Flash an ESP32 with `econet_heatpump_water_heater.yaml` from `github://esphome-econet/esphome-econet@main`, wired to a Rheem heat pump water heater, and enable the (disabled-by-default) `econet_climate` entity in HA.
2. `esphome logs <device>.yaml` on ESPHome 2026.4.0.
3. Observe `ERROR Fatal error: protocol.data_received() call failed` + disconnect every ~30 seconds starting with the first non-initial state publish.

kingpanther13:

Makes `_format_climate` and `_format_sensor` in `state_log_formatter.py` defensive about the concrete type of the `EntityInfo` they receive, so a state/info type mismatch falls back to the info-less output path instead of crashing the whole API connection with `AttributeError`.

Fixes #1588.

## The bug

`log_runner.subscribe_log_states_runner` builds its info lookup keyed only by entity key:

\`\`\`python
entity_info: dict[int, EntityInfo] = {e.key: e for e in entities}
\`\`\`

When two entities on the same device share an entity key hash — for example, an external component that exposes the same physical appliance as both a legacy `climate` entity and a newer `water_heater` entity (this is the reproducer, via the [`esphome-econet`](https://github.com/esphome-econet/esphome-econet) package on a Rheem heat pump water heater) — the later entry wins in that dict. So a `ClimateStateResponse` for key `K` gets paired with a `WaterHeaterInfo` instead of a `ClimateInfo`, dispatch routes `ClimateState → _format_climate`, and `_format_climate` crashes on `info.supports_two_point_target_temperature` (which exists on `ClimateInfo` only). The whole log-runner connection tears down and reconnects, only to crash again on the next state publish.

The same latent issue applies to `_format_sensor`, which reads `info.accuracy_decimals` / `info.unit_of_measurement` without checking that `info` is a `SensorInfo`.

## The fix

Inside each formatter that reads info-type-specific attributes, narrow `info` via `isinstance` to the expected subclass; fall through to the info-less output path otherwise. Name rendering via `_name(info)` is untouched and still works on any `EntityInfo`, so log lines still show the right entity name even when the info type doesn't match the state type. This is strictly additive: every existing test (including `test_water_heater_basic`, which deliberately passes a `BinarySensorInfo` to exercise the cross-cutting-field path) continues to produce identical output.

Two regression tests added:
- `test_climate_mismatched_info_type` — pairs `ClimateState` with a `WaterHeaterInfo`, asserts no crash and that climate output is still produced.
- `test_sensor_mismatched_info_type` — pairs `SensorState` with a `BinarySensorInfo`, asserts the formatter falls back to the unit-less output path.

## Types of changes

- [ ] Other

- fixes #1588

- N/A (no `api.proto` change)

## Checklist

- [x] If api.proto was modified, a linked pull request has been made to [esphome](https://github.com/esphome/esphome) with the same changes. *(N/A — no proto change)*

Note on local testing: my dev environment is Termux on aarch64-android, where the `aioesphomeapi` install flow fails because its build deps include a Rust toolchain that has no `aarch64-unknown-linux-android` target. I ran `ruff check` and `ruff format --check` on both modified files (clean) and traced the logic by hand, but couldn't execute pytest locally. CI on this PR will be the first full test run — happy to push follow-ups if anything fails there.

---

*Disclosure: this PR was drafted with assistance from [Claude Code](https://claude.com/claude-code) (Anthropic's coding CLI). I reviewed the root-cause analysis, the fix, and the tests before opening the PR, and the commit is co-authored accordingly.*

### DIFF
--- a/aioesphomeapi/state_log_formatter.py
+++ b/aioesphomeapi/state_log_formatter.py
@@ -77,9 +77,12 @@ def _format_sensor(state: SensorState, info: SensorInfo | None) -> str | None:
     if state.missing_state:
         return None
     name = _name(info)
-    if info is not None:
-        decimals = max(0, info.accuracy_decimals)
-        unit = info.unit_of_measurement
+    # Narrow to the expected info type so a mismatched EntityInfo (e.g. from
+    # two entities sharing a key hash) can't crash on type-specific attributes.
+    sensor_info = info if isinstance(info, SensorInfo) else None
+    if sensor_info is not None:
+        decimals = max(0, sensor_info.accuracy_decimals)
+        unit = sensor_info.unit_of_measurement
         return f"[S][sensor]: '{name}' >> {state.state:.{decimals}f} {unit}"
     return f"[S][sensor]: '{name}' >> {state.state}"
 
@@ -207,6 +210,9 @@ def _format_climate(state: ClimateState, info: ClimateInfo | None) -> str | None
         f"[S][{tag}]: '{_name(info)}' >>",
         _detail(tag, "Mode", _enum_name(state.mode)),
     ]
+    # Narrow to ClimateInfo so a mismatched EntityInfo (e.g. from two entities
+    # sharing a key hash) can't crash on type-specific attributes.
+    climate_info = info if isinstance(info, ClimateInfo) else None
     if state.action is not None and state.action != ClimateAction.OFF:
         parts.append(_detail(tag, "Action", state.action.name))
     if state.fan_mode is not None:
@@ -222,7 +228,7 @@ def _format_climate(state: ClimateState, info: ClimateInfo | None) -> str | None
     ct = state.current_temperature
     if not isnan(ct):
         parts.append(_detail(tag, "Current Temperature", f"{ct:.2f}°C"))
-    if info and info.supports_two_point_target_temperature:
+    if climate_info and climate_info.supports_two_point_target_temperature:
         ttl = state.target_temperature_low
         tth = state.target_temperature_high
         if not isnan(ttl) and not isnan(tth):
@@ -237,9 +243,17 @@ def _format_climate(state: ClimateState, info: ClimateInfo | None) -> str | None
         tt = state.target_temperature
         if not isnan(tt):
             parts.append(_detail(tag, "Target Temperature", f"{tt:.2f}°C"))
-    if info and info.supports_current_humidity and not isnan(state.current_humidity):
+    if (
+        climate_info
+        and climate_info.supports_current_humidity
+        and not isnan(state.current_humidity)
+    ):
         parts.append(_detail(tag, "Current Humidity", f"{state.current_humidity:.0f}%"))
-    if info and info.supports_target_humidity and not isnan(state.target_humidity):
+    if (
+        climate_info
+        and climate_info.supports_target_humidity
+        and not isnan(state.target_humidity)
+    ):
         parts.append(_detail(tag, "Target Humidity", f"{state.target_humidity:.0f}%"))
     return "\n".join(parts)
 
--- a/tests/test_state_log_formatter.py
+++ b/tests/test_state_log_formatter.py
@@ -38,6 +38,7 @@ from aioesphomeapi.model import (
     UpdateState,
     ValveOperation,
     ValveState,
+    WaterHeaterInfo,
     WaterHeaterMode,
     WaterHeaterState,
 )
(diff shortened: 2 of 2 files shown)


bdraco:
If two entities are sharing the same key, the ESPHome config is fundamentally broken, and this PR is papering over the issue.

bdraco (changes requested):
Thanks for the detailed writeup and repro, the tracedown is really helpful.

I don't want to merge this as is though. As an ESPHome maintainer, my read is that this is papering over a broken config rather than fixing the real problem. ESPHome already enforces duplicate name/object_id checks at compile time, so before we add client side defenses I need to understand how the econet external component is managing to ship two entities on one device with a colliding key in the first place. That's the thread I want to pull on, because if it's slipping past the existing validation, the fix is to close that hole in ESPHome so this class of collision can't compile, not to teach the client to tolerate it.

Hiding it in the log formatter means the logs will silently render a WaterHeaterInfo as climate output instead of surfacing the collision, which makes the underlying bug harder to find rather than easier. It also only covers two formatters; every other formatter that reads type specific attributes has the same latent crash, so this becomes whack a mole as soon as the next collision lands on a different entity type.

If we do end up wanting a defensive layer in aioesphomeapi on top of the real fix, I'd rather it live in one place, either `log_runner.on_state` checking that `type(state)` is compatible with `type(info)` before dispatch, or `format_state_log` passing `info=None` when the pair doesn't match. And `log_runner` building `{e.key: e for e in entities}` could detect the collision and log a clear warning naming both entities, so on older firmware the user sees the real problem instead of a confusing traceback or silently wrong logs.

Going to dig into the econet component first to see what it's actually doing before we decide where the fix belongs.

bdraco:
Thanks for flagging this and for the thorough root cause writeup, it was genuinely helpful in tracking down where the real bug lives.

After digging in, I don't want to change anything on the ESPHome side; collisions across different entity types on the same device are fine, the client just has to handle them. The actual bug is in `aioesphomeapi/log_runner.py`, where the info lookup is flattened to `{e.key: e for e in entities}` with no type dimension, so a later entity of a different type overwrites an earlier one. Home Assistant already keys its info storage by type and doesn't hit this, so the fix is to make log_runner match that pattern rather than guard each formatter.

I'll take it from here and put up a PR on the log_runner side, so I'll close this one out. Really appreciate you bringing it to our attention.

kingpanther13:
Thank you for taking over! And to be clear I have no affiliation with econet repo so sorry if I overstepped, I filed an issue on their repo to make them aware of the issue too. I'm just a guy that uses the econet repo, and I purposefully kept it on the old climate entity instead of switching to water heater since it would be a headache to change all of my automations.

PR #1645, opened 2026-05-15
Reject msg_type=0 in process_packet to prevent silent last-class fallback

### LINKED ISSUE #1644: Security: Negative-index lookup in `process_packet` causes msg_type=0 to be parsed as the last protobuf type instead of being rejected
## Problem

`MESSAGE_NUMBER_TO_PROTO` is a Python `tuple` indexed as `MESSAGE_NUMBER_TO_PROTO[msg_type_proto - 1]`. When a peer sends `msg_type_proto == 0` (a value that is not assigned to any real message in `MESSAGE_TYPE_TO_PROTO`), the expression evaluates to `MESSAGE_NUMBER_TO_PROTO[-1]`, which silently returns the *last* registered protobuf class rather than raising `IndexError`. The function then attempts to `MergeFromString` the attacker-controlled payload into that arbitrary class. The `IndexError`-handling branch exists specifically to silently drop unknown message types (`if isinstance(e, IndexError): … return`), but it is bypassed for `msg_type=0` because `-1` is a valid index for non-empty tuples. The "last" message type rotates over time as new entries are appended in `core.py`, so the unintended target class is whatever happens to be last in the registry today (currently `ListEntitiesRadioFrequencyResponse`).

## Why This Matters

A peer can route arbitrary bytes into a deserializer it should never have been able to address, against a class chosen by the order of additions to the registry rather than by the sender. Realistically this just causes `MergeFromString` to fail and `report_fatal_error()` to drop the connection, but it bypasses the intended "unknown message type → silently ignore" policy and exposes a parser the developer never meant to expose to type=0 input. It also creates a footgun: future refactors that move a security-sensitive message to the end of the registry would suddenly become reachable via msg_type=0.

## Suggested Fix

Reject non-positive message types explicitly before indexing, e.g.:
```python
def process_packet(self, msg_type_proto, data):
    if msg_type_proto < 1 or msg_type_proto > len(MESSAGE_NUMBER_TO_PROTO):
        if self._debug_enabled:
            _LOGGER.debug("%s: Skipping unknown message type %s",
                          self.log_name, msg_type_proto)
        return
    klass, merge = MESSAGE_NUMBER_TO_PROTO[msg_type_proto - 1]
    ...
```
Alternatively, store the registry as a `dict[int, …]` and use `.get(msg_type_proto)`.

## Details

| | |
|---|---|
| **Severity** | 🟢 Low |
| **Category** | request_handling |
| **Location** | `aioesphomeapi/connection.py:1024-1059 (`APIConnection.process_packet`) and connection.py:66-68 (`MESSAGE_NUMBER_TO_PROTO`)` |
| **Effort** | ⚡ Quick fix |

---

bdraco:

`MESSAGE_NUMBER_TO_PROTO[msg_type_proto - 1]` with `msg_type_proto == 0` evaluates to `MESSAGE_NUMBER_TO_PROTO[-1]`, which silently returns the *last* registered protobuf class instead of raising `IndexError`. The `IndexError`-handling branch in `process_packet` exists specifically to drop unknown message types, but it is bypassed for `msg_type=0`, so the attacker-controlled payload is routed to whatever class happens to be last in the registry today (currently `ListEntitiesRadioFrequencyResponse`).

This adds an explicit guard rejecting non-positive message types up front so the "unknown message type → silently ignore" policy applies uniformly. The check is one branch, perfectly predicted in the common case (`msg_type >= 1`), so it preserves the hot-path performance the existing comment calls out.

I confirmed the bug empirically before fixing: with the original code, a `msg_type=0, payload=b""` packet is parsed as `ListEntitiesRadioFrequencyResponse` rather than being rejected. The new regression test (`test_zero_protobuf_message_type_rejected`) fails on the unfixed code and passes with the fix.

## Types of changes

- [ ] Other

- fixes https://github.com/esphome/aioesphomeapi/issues/1644

- N/A (client-side only)

## Checklist:

### DIFF
--- a/aioesphomeapi/connection.pxd
+++ b/aioesphomeapi/connection.pxd
@@ -73,6 +73,7 @@ cdef object _handle_timeout
 cdef object _handle_complex_message
 
 cdef tuple MESSAGE_NUMBER_TO_PROTO
+cdef Py_ssize_t _MESSAGE_NUMBER_TO_PROTO_LEN
 
 
 @cython.dataclasses.dataclass
--- a/aioesphomeapi/connection.py
+++ b/aioesphomeapi/connection.py
@@ -66,6 +66,7 @@ _LOGGER = logging.getLogger(__name__)
 MESSAGE_NUMBER_TO_PROTO: tuple[
     tuple[Callable[[], message.Message], Callable[[message.Message, bytes], None]], ...
 ] = tuple((msg, msg.MergeFromString) for msg in MESSAGE_TYPE_TO_PROTO.values())
+_MESSAGE_NUMBER_TO_PROTO_LEN = len(MESSAGE_NUMBER_TO_PROTO)
 
 
 PREFERRED_BUFFER_SIZE = 2097152  # Set buffer limit to 2MB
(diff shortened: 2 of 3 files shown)


bdraco:
Addressed in 99d82f3 — extended the guard to also reject `msg_type_proto > _MESSAGE_NUMBER_TO_PROTO_LEN` (precomputed module-level len so it's a single comparison, not a `len()` call per packet). The dead `IndexError` branch in the `except` is now removed since both edges are caught up front. Existing `test_unknown_protobuf_message_type_logged` (msg_type=16385) already covers the upper-bound path through the new guard.

PR #1648, opened 2026-05-15
Bound camera image stream reassembly to prevent memory exhaustion

### LINKED ISSUE #1643: Security: Camera image stream accumulates attacker-controlled chunks without bound (memory exhaustion)
## Problem

`on_state_msg` reassembles `CameraImageResponse` chunks in a per-subscription `image_stream: dict[int, list[bytes]]`. The dictionary key is `cam_msg.key`, and the buffered bytes are released only when `cam_msg.done` is True. Both `key` and `done` are fields of a protobuf message received from the peer, so the peer fully controls whether the buffer is ever drained. A malicious or compromised device (or, if encryption is disabled, an unauthenticated network attacker that completes the trivial plaintext handshake) can:
1. Stream `CameraImageResponse` packets with `done=False` indefinitely, growing `image_stream[k]` without bound, and/or
2. Vary `cam_msg.key` (a 32-bit field) to allocate up to ~4 billion separate list entries in the dict.
There is no per-stream byte limit, no per-key limit, no overall cap, and no timeout-based eviction. The `subscribe_states` path is enabled by default by every Home Assistant ESPHome integration that has a camera (and by `aioesphomeapi-logs` whenever `--no-states` is not passed), so the attack surface is wide.

## Why This Matters

A single misbehaving device can OOM the host process of any client subscribed to its states — most notably Home Assistant. This is a remote memory-exhaustion DoS that requires only the ability to send protobuf packets over an existing API session, and trivially escalates to availability loss for the entire smart-home control plane.

## Suggested Fix

Enforce per-stream and per-key limits in `on_state_msg`. For example, track the cumulative size for each `key` and abort if it exceeds a sane camera frame ceiling (e.g. 8 MiB), drop and log if a single client accumulates more than N concurrent in-flight `key`s, and add a watchdog that evicts streams that have not received a `done=True` within a few seconds. Sketch:
```python
MAX_CAMERA_FRAME_BYTES = 8 * 1024 * 1024
MAX_INFLIGHT_KEYS = 4

if msg_type is CameraImageResponse:
    cam_msg = msg
    parts = image_stream.get(cam_msg.key)
    if parts is None:
        if len(image_stream) >= MAX_INFLIGHT_KEYS:
            _LOGGER.warning("Dropping camera stream: too many in-flight keys")
            return
        parts = []
        image_stream[cam_msg.key] = parts
    parts.append(cam_msg.data)
    if sum(map(len, parts)) > MAX_CAMERA_FRAME_BYTES:
        del image_stream[cam_msg.key]
        _LOGGER.warning("Dropping camera stream: frame too large")
        return
    if cam_msg.done:
        ...
```

## Details

| | |
|---|---|
| **Severity** | 🟡 Medium |
| **Category** | dos |
| **Location** | `aioesphomeapi/client_base.py:68-94 (state-message dispatcher used by `subscribe_states`)` |
| **Effort** | ⚡ Quick fix |

---

bdraco:

`on_state_msg` reassembles `CameraImageResponse` chunks in a per-subscription `image_stream` dict keyed by `cam_msg.key`. Both `cam_msg.key` and `cam_msg.done` are peer-controlled fields, so without limits a single misbehaving or compromised device can:

1. Stream chunks with `done=False` indefinitely, growing one entry in `image_stream` without bound, **or**
2. Vary `cam_msg.key` (a 32-bit field) to allocate up to ~4 billion separate buffers.

Either path OOMs the host process of every state subscriber — most notably Home Assistant, which subscribes whenever an ESPHome integration has a camera.

This caps the per-stream byte total at **8 MiB** (generous headroom over the largest realistic ESP32-S3 5MP JPEG) and the number of concurrent in-flight keys at **4**. When either bound is exceeded the partial buffer is dropped and a warning is logged, so a later `done=True` for the same key cannot flush an oversized frame.

Also switches the accumulator from `list[bytes]` to `bytearray` so the per-chunk size check is O(1) instead of summing the list every chunk — without that, an attacker could force quadratic work by sending many small chunks just under the cap.

## Types of changes

- [ ] Other

- fixes https://github.com/esphome/aioesphomeapi/issues/1643

- N/A (client-side only)

## Checklist:

### DIFF
--- a/aioesphomeapi/client_base.py
+++ b/aioesphomeapi/client_base.py
@@ -64,10 +64,17 @@ _LOGGER = logging.getLogger(__name__)
 # connection is poor.
 KEEP_ALIVE_FREQUENCY = 20.0
 
+# Caps on the per-subscription camera reassembly buffer. The peer fully
+# controls cam_msg.key and cam_msg.done, so without these limits a single
+# device can stream chunks indefinitely (memory exhaustion DoS) or rotate
+# keys to allocate ~4 billion list entries.
+MAX_CAMERA_FRAME_BYTES = 8 * 1024 * 1024
+MAX_INFLIGHT_CAMERA_KEYS = 4
+
 
 def on_state_msg(
     on_state: Callable[[EntityState], None],
-    image_stream: dict[int, list[bytes]],
+    image_stream: dict[int, bytearray | None],
     msg: message.Message,
 ) -> None:
     """Handle a state message."""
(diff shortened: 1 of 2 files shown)


bdraco:
Addressed in b4b1bbd — discarded keys are now tracked with a `None` sentinel in `image_stream` instead of being deleted. Subsequent chunks for a discarded key are silently ignored, and a later `done=True` clears the entry without emitting a `CameraState` (so a peer can't flush a truncated frame after we drop the buffer). The dict size is still bounded by `MAX_INFLIGHT_CAMERA_KEYS` because the discarded entry occupies the slot until `done=True` arrives. Test updated to assert no truncated frame is delivered for the dropped key, and that a fresh stream on the same key works after the discarded entry is cleared.

bdraco:
Addressed in ca6bbeb — moved the size check to use `len(buf) + len(cam_msg.data) > MAX_CAMERA_FRAME_BYTES` and short-circuit before `buf.extend`, so an oversized chunk never gets copied into the reassembly buffer in the first place. The bound is now strict rather than "strict after one chunk's worth of overshoot".

bdraco:
@bluetoothbot review

bdraco:
Addressed in 6f1214e — when an oversized chunk also has `done=True`, the entry is now `del`'d instead of being marked `None`. The tombstone is only useful for partial streams that might continue; for completed oversized frames there's nothing left to ignore. Otherwise a peer could send 4 single-message oversized done=True frames with distinct keys to fill `image_stream` with tombstones and DoS all subsequent camera traffic. New regression test (`test_subscribe_states_camera_oversized_done_chunk_does_not_tombstone`) — verified it fails on the previous commit.

PR #1651, opened 2026-05-15
Bound plaintext varuint length and frame size to prevent buffer-growth DoS

### LINKED ISSUE #1642: Security: Plaintext frame helper allows unbounded buffer growth and giant‑integer DoS via crafted varuint
## Problem

`APIPlaintextFrameHelper._read_varuint()` decodes a varuint from the receive buffer with no upper bound on the value or on the number of continuation bytes. Two related DoS primitives follow. (1) An attacker who can speak to a plaintext-mode client (anyone on the LAN if the device runs without `noise_psk`, or a MITM that wins the race against the real device) can stream bytes whose high bit is always set (`\x80\x80\x80…`). The varuint never terminates, so `_read_varuint` returns `-1`, `data_received` returns, and the next chunk re-scans from `self._pos = 0`. Because `_remove_from_buffer()` is never called in that path, `_add_to_buffer()` keeps doing `self._buffer += bytes_data` (base.py:91) — the buffer grows without bound and each scan rebuilds an ever-larger Python `int` (`result |= (val & 0x7F) << bitpos`), which is O(bits²) in memory for the bigint representation. (2) Even with a well-formed preamble, the `length` varuint is unbounded; sending e.g. a 4 GiB length causes `_read()` to keep buffering until that many bytes are received before it ever yields a frame. There is no `MAX_FRAME_SIZE` check anywhere in the plaintext path. The noise path is incidentally protected because its length field is the fixed 16-bit header (max 65535), but plaintext has no such cap.

## Why This Matters

Long‑running consumers of this library (notably Home Assistant integrations, log readers, and Bluetooth proxies) can be remotely OOM‑killed by anyone who can reach a plaintext-mode ESPHome device on the network. Crashing a Home Assistant integration is a real availability impact for smart-home users, and no authentication is required because plaintext mode has none. Devices without `noise_psk` are still common in deployed installs.

## Suggested Fix

Bound both the varuint and the resulting frame size. For example, in `_read_varuint` cap iterations at e.g. 5 bytes (the canonical 32‑bit varuint width) and treat anything longer as a protocol error that closes the connection; and in `data_received` validate `length` against a constant such as `MAX_PLAINTEXT_FRAME_SIZE = 2 * 1024 * 1024` (matching `PREFERRED_BUFFER_SIZE`) and call `_handle_error_and_close(ProtocolAPIError(...))` when exceeded. Sketch:
```python
MAX_VARUINT_BYTES = 5
MAX_FRAME_SIZE = 2 * 1024 * 1024

def _read_varuint(self):
    ...
    bytes_read = 0
    while self._buffer_len > self._pos and bytes_read < MAX_VARUINT_BYTES:
        ...
        bytes_read += 1
    if bytes_read == MAX_VARUINT_BYTES:
        raise ProtocolAPIError("varuint too long")
    return -1
```
And in `data_received`, after decoding `length`, reject it if `length > MAX_FRAME_SIZE`. Also consider giving `_add_to_buffer` an absolute cap that closes the connection rather than appending forever.

## Details

| | |
|---|---|
| **Severity** | 🟡 Medium |
| **Category** | dos |
| **Location** | `aioesphomeapi/_frame_helper/plain_text.py:32-76 (with aioesphomeapi/_frame_helper/base.py:73-92)` |
| **Effort** | ⚡ Quick fix |

---

bdraco:

`APIPlaintextFrameHelper._read_varuint()` decoded varuints with no upper bound on continuation bytes, and `data_received` had no cap on the length varuint. Two related DoS vectors followed for any peer that can speak to a plaintext-mode client (anyone on the LAN of a device running without `noise_psk`):

1. **Unbounded varuint** — stream `\x80\x80\x80…` forever. The varuint never terminates, `_read_varuint` returns `-1`, `_remove_from_buffer` is never called, so `_add_to_buffer` keeps appending and every chunk re-scans from `self._pos = 0`, rebuilding an ever-larger Python `int`. O(N²) work over the buffer size on top of unbounded memory.
2. **Unbounded frame length** — send a well-formed preamble plus a length varuint of e.g. 4 GiB. `_read()` keeps buffering until that many bytes arrive — easy OOM.

The noise path is incidentally protected because its length is a fixed 16-bit header (max 65535); plaintext had no equivalent cap.

This caps the varuint at **4 bytes** and the per-frame length at **65535**. Both bounds raise `ProtocolAPIError` and close the connection.

The 4-byte varuint cap is intentional — under Cython, `_read_varuint`'s `result` is `unsigned int` and the function returns `cdef int`. A 5-byte varuint can decode up to `0xFFFFFFFF`; values with bit 31 set would come back as a *negative* int and silently hit the sentinel branch in `data_received`, leaving the buffer-growth DoS open. 4 bytes keeps the decoded value in `[0, 2**28 - 1]`, well clear of the unsigned→signed cast boundary, and is well above any value the protocol actually uses (length is bounded at 65535 = 17 bits; msg_type fits in 1 byte).

`data_received` checks for the two explicit sentinels (`_VARUINT_INCOMPLETE` / `_VARUINT_PROTOCOL_ERROR`) rather than `< 0`, so the same trap can't be re-introduced by a future widening of the varuint cap without an accompanying caller change.

## Types of changes

- [ ] Other

- fixes https://github.com/esphome/aioesphomeapi/issues/1642

- N/A (client-side only)

## Checklist:

### DIFF
--- a/aioesphomeapi/_frame_helper/plain_text.pxd
+++ b/aioesphomeapi/_frame_helper/plain_text.pxd
@@ -6,6 +6,12 @@ from .packets cimport make_plain_text_packets
 
 cdef bint TYPE_CHECKING
 
+cdef int _MAX_VARUINT_BYTES
+cdef unsigned int _MAX_VARUINT_BITPOS
+cdef int _MAX_PLAINTEXT_FRAME_SIZE
+cdef int _VARUINT_INCOMPLETE
+cdef int _VARUINT_PROTOCOL_ERROR
+
 
 cdef class APIPlaintextFrameHelper(APIFrameHelper):
 
--- a/aioesphomeapi/_frame_helper/plain_text.pxd
+++ b/aioesphomeapi/_frame_helper/plain_text.pxd
@@ -19,6 +19,8 @@ cdef class APIPlaintextFrameHelper(APIFrameHelper):
 
     cdef void _error_on_incorrect_preamble(self, int preamble) except *
 
+    cdef void _close_on_oversized_varuint(self) except *
+
     @cython.locals(
         result="unsigned int",
         bitpos="unsigned int",
--- a/aioesphomeapi/_frame_helper/plain_text.py
+++ b/aioesphomeapi/_frame_helper/plain_text.py
@@ -55,10 +55,15 @@ class APIPlaintextFrameHelper(APIFrameHelper):
     def _read_varuint(self) -> _int:
         """Read a varuint from the buffer.
 
-        Returns the decoded value, or _VARUINT_INCOMPLETE if more bytes are
-        needed. If the varuint exceeds _MAX_VARUINT_BYTES the connection is
-        closed and _VARUINT_PROTOCOL_ERROR is returned, so the caller can
-        treat any negative value as "stop processing this frame".
+        Pure C path with no Python calls. Returns one of:
+          * the decoded value (>= 0), if a complete varuint was read;
+          * _VARUINT_INCOMPLETE, if the buffer ran out mid-varuint;
+          * _VARUINT_PROTOCOL_ERROR, if the varuint exceeds
+            _MAX_VARUINT_BYTES (the caller is responsible for closing the
+            connection).
+
+        Callers must check for both sentinels explicitly rather than
+        treating any negative value generically.
         """
         if TYPE_CHECKING:
             assert self._buffer is not None, "Buffer should be set"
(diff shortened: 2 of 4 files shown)


bdraco:
@bluetoothbot review

bdraco:
Addressed both Copilot comments in d6f5ccf — the underlying concern was that with `result="unsigned int"` (32-bit) and `cdef int` return, a 5-byte varuint encoding a value in [2³¹, 2³²) would cast to a *negative* int and silently hit the new `< 0` sentinel branch in `data_received`, leaving the buffer-growth DoS open.

Capped at **4 bytes** instead of 5 — max decoded value is now 2²⁸ - 1, which always fits in a signed 32-bit int. 4 bytes is well above any value the protocol uses (length is bounded at 65535 = 17 bits by the new frame-size cap; msg_type fits in 1 byte). Existing overlong-varuint tests still cover the rejection path (they send 6 continuation bytes; rejection now fires at the 5th rather than the 6th).

bdraco:
Applied Copilot's defensive structure in 12bf906. `data_received` now explicitly checks for the two known sentinels (`_VARUINT_INCOMPLETE` / `_VARUINT_PROTOCOL_ERROR`) and treats any other negative as a protocol error that closes the connection — so even if a future change re-introduces an unsigned->signed cast bug in `_read_varuint`, the buffer can't silently grow. Also switched `_error_on_incorrect_preamble`'s format from `:02x` to `!r` so it doesn't crash if it ever does see a negative.

bdraco:
@bluetoothbot review

bdraco:
@bluetoothbot — addressed the remaining concerns in 7328cc5:

**Suggestion 2 (mirror constant)**: `MAX_PLAINTEXT_FRAME_SIZE` is now the Python-importable name and `_MAX_PLAINTEXT_FRAME_SIZE = MAX_PLAINTEXT_FRAME_SIZE` is the cdef-int alias used internally per the .pxd. Tests import `MAX_PLAINTEXT_FRAME_SIZE` directly, so they can't drift from the source of truth.

**Blocking — regression test**: added `test_plaintext_frame_helper_rejects_5_byte_high_bit_varuint` that sends `PREAMBLE + b"\xff\xff\xff\xff\x0f"` and asserts the connection closes with ProtocolAPIError. With the 4-byte cap the rejection fires at the 5th byte (bitpos limit) before the unsigned->signed cast can happen, but this pins the guarantee.

**Important — PR description**: updated to drop the stale claim about exception propagation and explain the 4-byte cap (vs. 5-byte) is intentional to stay clear of the unsigned->signed cast boundary.

bdraco:
@bluetoothbot review

bdraco:
Addressed all three Copilot comments in 948c6b7:

**Line 84 — _handle_error_and_close inside noexcept _read_varuint**: moved the close call into a new `_close_on_oversized_varuint` helper called from `data_received`. `_read_varuint` is now a pure C path (sentinel returns, no Python calls), so the `noexcept` declaration in the .pxd is honest — there is no longer any code that could raise.

**Line 61 — docstring**: rewrote to enumerate the two explicit sentinels and discourage generic `< 0` checks at call sites (which is exactly the trap an unsigned->signed cast regression in `result` type would re-open).

**Line 145 — preamble format**: switched from `{preamble!r}` (decimal) back to `0x{preamble:02x}` so protocol captures show the byte-oriented hex form. Safe because the sentinel checks in `data_received` filter out negatives before this code is reached.

bdraco:
@bluetoothbot review

PR #1656, opened 2026-05-15
Sanitize unauthenticated noise hello fields before logging

### LINKED ISSUE #1655: Security: Unauthenticated server-supplied name/MAC/handshake explanation is interpolated into log messages (log injection)
## Problem

During the Noise hello phase, `_handle_hello()` reads `server_name` (line 194) and `mac_address` (line 209) from raw bytes the peer sent **before** the PSK-authenticated handshake completes. These attacker-controllable strings are then (a) interpolated unescaped into log/error messages — `f"... Server sent a different name '{server_name}'"` (line 200), `f"... Server sent a different mac '{mac_address}'"` (line 214) — and (b) stashed in `self._server_name` / `self._server_mac` for reuse by every later error (`_decode_noise_psk`, `_error_on_incorrect_preamble`, `_handle_frame`'s `EncryptionErrorAPIError`). `_error_on_incorrect_preamble` (line 259) similarly takes attacker bytes (`msg[1:].decode()`) and embeds them in `f"Handshake failure: {explanation}"`. An on-path attacker who can intercept the TCP connection (no PSK needed — this happens before the handshake) can inject newlines, ANSI escape sequences, and forged log lines into operator-visible logs and Telegram/HA notification streams; with control characters they can clear the screen, hijack TTYs, or smuggle `Co-Authored-By:` / fake stack traces into incident reports.

## Why This Matters

ESPHome deployments typically run on the LAN with permissive routing, where ARP/DHCP spoofing or a compromised VLAN device puts an attacker on-path. The injected text reaches Home Assistant's UI log viewer, syslog, and any monitoring webhook unfiltered, which can deceive operators (impersonate a different device's identity in alert messages) and exploit terminal emulators that interpret CSI sequences.

## Suggested Fix

Sanitize the unauthenticated fields before storing or formatting. At minimum strip non-printable bytes and cap length, e.g.:
```python
def _safe_label(raw: str, *, limit: int = 64) -> str:
    cleaned = "".join(c for c in raw if c.isprintable() and c not in "\r\n")
    return cleaned[:limit]

server_name = _safe_label(server_hello[1:server_name_i].decode("utf-8", "replace"))
```
Apply the same sanitizer to `mac_address` and to `explanation` in `_error_on_incorrect_preamble`. Consider also rejecting non-hex MACs outright since the wire format is fixed.

## Details

| | |
|---|---|
| **Severity** | 🟢 Low |
| **Category** | request_handling |
| **Location** | `aioesphomeapi/_frame_helper/noise.py:191-219, 257-263` |
| **Effort** | ⚡ Quick fix |

---

bdraco:

The noise hello phase exchanges `server_name`, `mac_address`, and a handshake-failure `explanation` as raw bytes **before** the PSK-authenticated handshake completes. The previous code interpolated those attacker-controlled strings unescaped into log/error messages (`f"... Server sent a different name '{server_name}'"`) and stored them on `self` for reuse by every later error path (`_decode_noise_psk`, `_error_on_incorrect_preamble`, `EncryptionErrorAPIError`).

An on-path attacker (ARP spoof, compromised VLAN device) could inject CRLF + ANSI escape sequences into operator-visible logs (HA UI, syslog, monitoring webhooks) — forging fake log lines, hijacking terminals via CSI sequences, and impersonating a different device's identity in alert messages.

This adds a `_safe_label()` helper that decodes with `errors="replace"` and strips any non-printable character, then caps the length to the firmware's actual wire-format limits (verified against `esphome/core/{config.py,helpers.h,entity_base.h}` and `api_frame_helper_noise.cpp`):

- **name**: 32 (firmware `ESPHOME_DEVICE_NAME_MAX_LEN = 31` from `validate_hostname`; allowed chars are `[a-z0-9_-]`)
- **mac**: 16 (firmware `MAC_ADDRESS_BUFFER_SIZE - 1 = 12` lowercase hex chars from `format_hex_to`)
- **explanation**: 64 (firmware reject buffer is 32 bytes minus the 1-byte failure code)

Margins on each let benign forward-compat tweaks through without breaking clients.

The constants and helper are `cdef`-typed in the `.pxd` for hot-path performance; `MAX_NAME_LEN` / `MAX_MAC_LEN` / `MAX_EXPLANATION_LEN` are re-exported as Python-importable aliases so tests can reference them, following the same pattern as `MAX_PLAINTEXT_FRAME_SIZE` in `plain_text.py`.

## Types of changes

- [ ] Other

- fixes https://github.com/esphome/aioesphomeapi/issues/1655

- N/A (client-side only)

## Checklist:

### DIFF
--- a/aioesphomeapi/_frame_helper/noise.pxd
+++ b/aioesphomeapi/_frame_helper/noise.pxd
@@ -16,6 +16,12 @@ cdef bytes NOISE_HELLO
 cdef object InvalidTag
 cdef object ESPHOME_NOISE_BACKEND
 
+cdef int _MAX_NAME_LEN
+cdef int _MAX_MAC_LEN
+cdef int _MAX_EXPLANATION_LEN
+
+cdef str _safe_label(bytes raw, int limit)
+
 cdef class APINoiseFrameHelper(APIFrameHelper):
 
     cdef object _noise_psk
--- a/aioesphomeapi/_frame_helper/noise.pxd
+++ b/aioesphomeapi/_frame_helper/noise.pxd
@@ -20,7 +20,7 @@ cdef int _MAX_NAME_LEN
 cdef int _MAX_MAC_LEN
 cdef int _MAX_EXPLANATION_LEN
 
-cdef str _safe_label(bytes raw, int limit)
+cdef str _safe_label_str(str raw, int limit)
 
 cdef class APINoiseFrameHelper(APIFrameHelper):
 
@@ -55,7 +55,9 @@ cdef class APINoiseFrameHelper(APIFrameHelper):
         server_name_i=int,
         mac_address_i=int,
         mac_address=str,
+        mac_address_raw=str,
         server_name=str,
+        server_name_raw=str,
     )
     cdef void _handle_hello(self, bytes server_hello) except *
 
--- a/aioesphomeapi/_frame_helper/noise.py
+++ b/aioesphomeapi/_frame_helper/noise.py
@@ -44,6 +44,34 @@ NOISE_HELLO = b"\x01\x00\x00"
 
 int_ = int
 
+# Caps match the firmware's actual wire-format limits:
+#   - name: ESPHOME_DEVICE_NAME_MAX_LEN = 31 (validate_hostname in core/config.py)
+#   - mac: MAC_ADDRESS_BUFFER_SIZE - 1 = 12 (lowercase hex, no separator)
+#   - explanation: 32-byte handshake-reject buffer minus the 1-byte failure code
+# A small extra margin on each lets benign forward-compat tweaks (e.g. firmware
+# bumping the max name length by a few chars) through without breaking clients.
+# MAX_* are the Python-importable forms (used by tests); _MAX_* are the cdef
+# int aliases used internally per the .pxd.
+MAX_NAME_LEN = 32
+MAX_MAC_LEN = 16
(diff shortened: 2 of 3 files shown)


bdraco:
@bluetoothbot review

bdraco:
Addressed all three Copilot comments in fa793f3:

**Lines 225, 237 — name/mac sanitization bypass**: now decoding raw once at the call site, comparing `expected_name`/`expected_mac` against the raw decoded value, and using the sanitized form only for storage and log/error messages. A peer sending `servicetest\r\n` no longer slips past the expected-name check by virtue of the sanitizer stripping the trailing chars.

**Line 290 — handshake explanation classification bypass**: same fix in `_error_on_incorrect_preamble`. `explanation == "Handshake MAC failure"` now compares against the raw decoded string, so a peer sending `Handshake MAC failure\0extra` is classified as a generic `HandshakeAPIError` (the actual wire situation) instead of being mis-promoted to `InvalidEncryptionKeyAPIError`.

**Line 13 — import order**: ruff auto-formats this back to `MAX_NAME_LEN, APINoiseFrameHelper` (uppercase-first per project isort config — see the existing `MAX_PLAINTEXT_FRAME_SIZE, APIPlaintextFrameHelper` import in the same file). Project rule, not Copilot's preference.

Three new regression tests cover each bypass path. Verified to fail without the raw-vs-sanitized split.

bdraco:
@bluetoothbot review

PR #1661, opened 2026-05-16
Sanitize peer DeviceInfo.name before storing as log_name

bluetoothbot:
# What does this implement/fix?

Sanitizes the peer-supplied `DeviceInfo.name` at the single sink
(`APIClientBase._set_name_from_device`) before it becomes
`self.cached_name` and `self.log_name`. Same `_safe_label_str` +
`MAX_NAME_LEN` cap that #1656 introduced for the noise hello path.

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Code quality improvements to existing code or addition of tests
- [ ] Other

**Related issue or feature (if applicable):**

Companion to #1660, which sanitizes the pre-handshake plaintext hello
path. This PR closes the matching post-handshake hole that fires on
every `device_info()` / `device_info_and_list_entities()` call.

**Pull request in [esphome](https://github.com/esphome/esphome) (if applicable):**

- N/A — client-side only, no `api.proto` change.

## Why

Plaintext sessions never authenticate the peer. After the hello path
is sanitized, `DeviceInfoResponse.name` remains the next
peer-controlled string that lands in `self.log_name` (via
`_set_name_from_device(info.name)` in `client.py`). From there it
flows into the `%s: ...` formatting on every `_LOGGER.warning` /
`_LOGGER.debug` call in `connection.py`, which lets an on-path
attacker forge log lines, hijack terminals with ANSI escapes, or
truncate context with NULs. Defense in depth for Noise sessions
too — a compromised authenticated device still shouldn't be able
to forge log lines on the host.

## How

Single sink: `_set_name_from_device` now does
`_safe_label_str(str(name), MAX_NAME_LEN)` instead of `str(name)`.
Both call sites in `client.py` (`device_info`,
`device_info_and_list_entities`) and the `received_name` plumbing
from `finish_connection` go through this helper, so one change
covers all entry points. `str(...)` flattens esphome's `Estr`
subclass; the sanitizer's `"".join(c for c in raw if c.isprintable())`
returns a plain `str`, so the contract is preserved.

Legitimate firmware names are hostname-valid (ESPHome enforces the
31-char hostname rules at config time), so real device names pass
through `_safe_label_str` unchanged. The 32-byte cap from
`MAX_NAME_LEN` leaves a single-char margin for forward-compat
firmware tweaks.

[...]

PR #1660, opened 2026-05-16
Sanitize plaintext hello name/server_info before logs and exceptions

bluetoothbot:
# What does this implement/fix?

Closes #1659. Applies the existing noise-path log-injection defense to the
plaintext `HelloResponse` path. `resp.server_info` and `resp.name` were
flowing unsanitized into `_LOGGER.debug` and into the `BadNameAPIError`
message; that message reaches `report_fatal_error`'s `_LOGGER.warning`
via `_raise_fatal_connection_exception`, so a peer-controlled CRLF /
ANSI sequence could forge log lines in operator logs, HA UI, and any
downstream log aggregator.

**What:** sanitize peer-supplied `name` / `server_info` in
`APIConnection._process_hello_resp` before they reach logs, the
`BadNameAPIError` payload, or `self.received_name`.

**How:**
- Hoist `_safe_label_str` and `MAX_NAME_LEN` / `MAX_MAC_LEN` /
  `MAX_EXPLANATION_LEN` from `_frame_helper/noise.py` to
  `_frame_helper/base.py` so noise and plaintext share one impl.
- `noise.py` re-exports the constants for back-compat with tests
  (`tests/test__frame_helper.py` imports `MAX_NAME_LEN` from the
  noise module path).
- `_process_hello_resp` compares against the **raw** decoded name (so a
  peer can't sneak past `expected_name` by appending stripped bytes —
  same pattern as `_frame_helper/noise.py::_handle_hello`) and uses
  the sanitized value only for the debug log, the
  `BadNameAPIError` payload, and `self.received_name`.

**Testing:** three new tests in `tests/test_connection.py` cover
mismatched-name with CRLF/ANSI bytes (must be stripped from both
the exception message and `received_name`), 4 KiB oversize name
(must be length-capped to `MAX_NAME_LEN`), and matched-name happy
path with malicious `server_info` (must not reach `caplog`). Full
suite passes locally (1010 passed, 2 skipped). Cython source-level
compilation succeeds for the updated `.pxd`.

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Code quality improvements to existing code or addition of tests
- [ ] Other

**Related issue or feature (if applicable):**

- fixes #1659

**Pull request in [esphome](https://github.com/esphome/esphome) (if applicable):**

- N/A — client-side only; protocol unchanged.

[...]

PR #1663, opened 2026-05-17
Truncate peer payload in Invalid protobuf message log/error

bluetoothbot:

When a peer sends a malformed protobuf frame, `APIConnection.process_packet`
currently logs the full raw payload via `%s` and embeds `repr(data)` into the
resulting `ProtocolAPIError`. The plaintext frame helper caps frames at
65535 bytes (and noise uses a 16-bit length header — same absolute cap), so
one bad message can land ~256KB of escaped peer bytes per failed connection
in journald, HA diagnostic bundles, and log aggregators. The peer can
reconnect and repeat, turning a fatal-on-first-byte protocol error into a
log-flood vector.

## How

- New module constant `MAX_PROTOBUF_ERROR_DATA_BYTES = 64` in `connection.py`.
- `process_packet` now logs `len=` separately and truncates the embedded
  payload to the cap, appending a `...(+N bytes)` marker when truncated.
- Same truncation is applied to the `ProtocolAPIError` text.
- Existing `test_bad_protobuf_message_drops_connection` was extended to
  assert the new `len=` field; a new
  `test_bad_protobuf_message_truncates_payload_in_log` exercises a payload
  4× the cap and pins both the truncation marker and the absence of the
  full escaped sequence.

## Why

Same defense-in-depth posture as #1648 (camera frame bytes), #1651
(plaintext varuint bound), and the recent peer-label sanitization series
(#1656, #1660, #1661): assume the bytes a peer can put into the receiving
side eventually hit logs, diagnostics, and exception traces — bound them
before they get there.

## Types of changes

- [ ] Other

- n/a — defense-in-depth follow-up to #1648 / #1651 / #1656 / #1660 / #1661.

- n/a — client-side log/error hardening only; no `api.proto` change.

## Checklist:

---
### Quality Report

**Changes**: 2 files changed, 62 insertions(+), 3 deletions(-)

**Code scan**: clean

**Tests**: failed (FAILED)

**Branch hygiene**: clean

*Generated by Kōan post-mission quality pipeline*


aioesphomeapi/connection.py
```diff
@@ -115,6 +115,14 @@
 
 WRITE_EXCEPTIONS = (RuntimeError, ConnectionResetError, OSError)
 
+# Cap on how many bytes of the offending peer payload we embed in the
+# "Invalid protobuf message" log entry and ProtocolAPIError text. A peer
+# can send up to MAX_PLAINTEXT_FRAME_SIZE (65535 bytes) or a 16-bit-length
+# noise frame of crafted garbage; with repr() escaping that balloons ~4x
+# in the log file. The truncated prefix is plenty for debugging the
+# message shape, and the total length is still reported separately.
+MAX_PROTOBUF_ERROR_DATA_BYTES = 64
```
bdraco, line 124: This needs to be a cdef

bdraco:
@bluetoothbot rr

aioesphomeapi/connection.py
bluetoothbot: Agreed in principle, but doing this naively will break the test. `tests/test_connection.py:31` imports `MAX_PROTOBUF_ERROR_DATA_BYTES` from `aioesphomeapi.connection`, and per the Cython gotchas in `CLAUDE.md`, declaring `cdef int MAX_PROTOBUF_ERROR_DATA_BYTES` in `connection.pxd` makes Cython treat the `.py` assignment as a C int store — the binding never lands in the Python module dict and the test's import raises `ImportError` under the production wheel build (pure-Python `SKIP_CYTHON=1` runs would still pass, hiding the regression).

The established pattern in this codebase for that situation is the dual-name shadow:

```python
# connection.py
MAX_PROTOBUF_ERROR_DATA_BYTES = 64
_MAX_PROTOBUF_ERROR_DATA_BYTES = MAX_PROTOBUF_ERROR_DATA_BYTES
```

```cython
# connection.pxd
cdef int _MAX_PROTOBUF_ERROR_DATA_BYTES
```

Then `process_packet` uses `_MAX_PROTOBUF_ERROR_DATA_BYTES` for the slice/compare and the public name stays importable for tests and external callers. Will push that fix.

Worth noting: the constant is only touched inside the `except Exception as e:` branch of `process_packet` (cold path — only when decode has already raised and the connection is about to be torn down), so there's no measurable PyLong-conversion cost on the hot wire-format path. Happy to apply the cdef anyway for consistency with the rest of the module's constants — just flagging that this isn't a perf-sensitive site like `_MAX_FRAME_SIZE`.

bdraco: Yes but lets be consistant. Things get copied

bdraco (approved):
Clear. Seen this be a real problem before in prod. Fix is warranted but I never had the cycles to do it.

PR #1733, opened 2026-05-23
Add flag to suppress time updates by client

### LINKED ISSUE #14658: Displaying logs uses the same api as Home Assistant and sends the time zone which changes the time zone of the device
### The problem

Esphome meets quantum physics: observing the system changes its state.

Displaying logs (either through the dashboard or using `esphome logs` change the time zone of the observed device to the time zone of the client looking for the logs.

How to reproduce:
- Set a time zone (i.e. GMT+2) in the time component (see sample yaml)
- `docker run --rm -v "${PWD}":/config --network host ghcr.io/esphome/esphome logs device.yaml`

Since no time zone was set in docker (no `TZ` env) the time zone of `device.yaml` will temporarily revert to UTC

### Which version of ESPHome has the issue?

2025.6.4

### What type of installation are you using?

Docker

### What platform are you using?

ESP32

### Component causing the issue

time

### YAML Config

```yaml
time:
  - platform: homeassistant
    id: homeassistant_time
    timezone: "Europe/Zurich"
```

### Anything in the logs that might be useful for us?

```txt
[18:22:38.716][C][homeassistant.time:010]: Home Assistant Time
[18:22:38.716][C][time:028]: Timezone: 'UTC0'
[18:22:38.716][C][time:031]: Current time: 2026-03-09 18:22:39
```

### Additional information

Most certainly related to: https://github.com/esphome/esphome/issues/10817

This is quite sneaky because the problem is transient. At some point the time zone reverts to the intended value. Still, the back & forth can create a mess with schedules

clydebarrow:

Related to:

https://github.com/home-assistant/core/issues/171846
https://github.com/esphome/esphome/issues/14658
https://github.com/esphome/esphome/pull/16583

Adds a flag to APIClient that will cause the connection to ignore time sync requests. The default value maintains the existing behaviour so is non-breaking.

## Types of changes

- [ ] Other

- fixes https://github.com/esphome/esphome/issues/14658

## Checklist:

### DIFF
--- a/tests/test_provide_time.py
+++ b/tests/test_provide_time.py
@@ -47,7 +47,7 @@ async def _make_connected_conn(
     provide_time: bool,
     resolve_host,
     aiohappyeyeballs_start_connection,
-) -> tuple[APIConnection, asyncio.Transport, APIPlaintextFrameHelper, asyncio.Task]:
+) -> tuple[APIConnection, asyncio.Transport, APIPlaintextFrameHelper]:
     """Set up a plaintext-connected PatchableAPIConnection with provide_time set."""
     loop = asyncio.get_running_loop()
     transport = MagicMock()
--- a/tests/test_provide_time.py
+++ b/tests/test_provide_time.py
@@ -41,7 +41,9 @@ async def test_api_client_provide_time_default() -> None:
 
 async def test_api_client_provide_time_false() -> None:
     """provide_time=False should be stored on _params."""
-    cli = PatchableAPIClient(address="127.0.0.1", port=6052, password=None, provide_time=False)
+    cli = PatchableAPIClient(
+        address="127.0.0.1", port=6052, password=None, provide_time=False
+    )
     assert cli._params.provide_time is False
 
 
@@ -118,4 +120,3 @@ async def test_get_time_response_not_sent_when_provide_time_false(
         transport.writelines.assert_not_called()
     finally:
         conn.force_disconnect()
-
--- a/tests/test_provide_time.py
+++ b/tests/test_provide_time.py
@@ -10,7 +10,7 @@ import asyncio
 from dataclasses import replace
 from functools import partial
 from typing import TYPE_CHECKING
-from unittest.mock import MagicMock
+from unittest.mock import MagicMock, patch
 
 from aioesphomeapi.api_pb2 import (  # type: ignore[attr-defined]
     GetTimeRequest,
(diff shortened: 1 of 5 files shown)


bdraco:
Tests are failing. Looks like an update is needed

tests/test_provide_time.py
```diff
@@ -0,0 +1,89 @@
...
+    params = replace(get_mock_connection_params(), provide_time=True)
+    conn = PatchableAPIConnection(params, mock_on_stop, True, None)
+
+    registered_types: list[type] = []
+
+    def capture(callback, msg_types):
+        registered_types.extend(msg_types)
+
+    conn._add_message_callback_without_remove = capture  # type: ignore[method-assign]
+    conn._register_internal_message_handlers()
```
bdraco: cdef's can't be patched because they aren't python code. Cython turns them into C

This has to be tested end to end instead

bdraco:
@bluetoothbot review

bdraco (approved):
LGTM assuming the bot doesn't find something I missed

tests/test_provide_time.py
bluetoothbot: Agreed, and the current revision does take the end-to-end route — `_make_connected_conn` drives a real plaintext connection and feeds an actual `GetTimeRequest` frame through `mock_data_received`, rather than patching the `cpdef` handler (which, as you note, becomes C under Cython and isn't patchable). One follow-up so the test is meaningful against the shipped Cython build: the `provide_time=True` assertion currently only checks `transport.write.called or transport.writelines.called`, which just proves *some* write happened. Frames are sent via `writelines` (never `write`), and any write would satisfy it. Mirroring `test_get_time_request_timezone_handling` would make it actually verify a GetTimeResponse:

```python
assert transport.writelines.call_count == 1
raw = b"".join(transport.writelines.call_args[0][0])
resp = GetTimeResponse()
resp.ParseFromString(raw[3:])
```

The `provide_time=False` side is already solid with `assert_not_called()`.

tests/test_provide_time.py
```diff
@@ -0,0 +1,118 @@
...
+    with patch_create_connection(loop, transport, connected):
+        connect_task = asyncio.create_task(connect(conn, login=False))
+        await connected.wait()
+        send_plaintext_hello(conn._frame_helper)
+        await connect_task
+        return conn, transport, conn._frame_helper
+
+
+def patch_create_connection(loop, transport, connected):
+    return __import__("unittest.mock", fromlist=["patch"]).patch.object(
```
bdraco: This could be a bit cleaner and is inconsistant with the rest of the tests. Bot caught it above and had a suggestion.

PR #1749, opened 2026-05-25
Add explicit __all__ to model to stop leaking stdlib imports

esphbot:

The package `__init__` does `from .model import *`, but `model.py` had no
`__all__`. So `import *` re-exported **every** public name bound in the
module — including model.py's own stdlib/typing imports. These leaked into
the public namespace:

```python
import aioesphomeapi
aioesphomeapi.field   # dataclasses.field
aioesphomeapi.UUID    # uuid.UUID
aioesphomeapi.partial # functools.partial
# ...also dataclass, lru_cache, asdict, cache, cast, fields, Any, ...
```

**What changed**

- Add `__all__` to `model.py`, limited to the names the module defines
  itself (classes, enums, helpers). `import *` now exports only those — the
  leaked stdlib/typing names are no longer attributes of `aioesphomeapi`.
- `fix_float_single_double_conversion` is a genuine `util` export (it is in
  `util.__all__`) that was only reachable through this leak, so it is now
  re-exported explicitly from `.util` in `__init__` to keep it public.
- A test pins `__all__` to model's own public definitions in **both**
  directions: a newly added public class must be added to `__all__` (or it
  silently stops being re-exported), and no imported name may appear in it.

Every previously-intended export is preserved (verified against the full
set of names defined in `model.py`); the only names removed from
`aioesphomeapi.*` are pure stdlib/typing imports that were never API.

## Types of changes

- [ ] Other

- n/a

- n/a

## Checklist:
  - [x] If api.proto was modified, a linked pull request has been made to [esphome](https://github.com/esphome/esphome) with the same changes. (api.proto not modified)


aioesphomeapi/__init__.py
```diff
@@ -27,3 +27,4 @@
 from .log_parser import LogParser, parse_log_message
 from .model import *
 from .reconnect_logic import ReconnectLogic
+from .util import fix_float_single_double_conversion
```
bdraco: I don't think `fix_float_single_double_conversion` was intended to be public API

bdraco:
@esphbot rr

aioesphomeapi/__init__.py
esphbot: Agreed — it shouldn't be public, and re-exporting it actually contradicts the goal of this PR. `fix_float_single_double_conversion` only reached `aioesphomeapi.*` because `model.py` imports it as a converter for its float fields and the un-`__all__`'d `from .model import *` re-exported it. `__init__.py` never imported it explicitly and never does `from .util import *`, so its presence in `util.__all__` only made it public within the `util` namespace, not at the package root. Keeping it is the one place this PR preserves a leak instead of closing it.

I'll drop the `from .util import fix_float_single_double_conversion` line. It's safe: `model.py` keeps its own internal import (the float converters still work), `tests/test_util.py` accesses it via `util.fix_float_single_double_conversion`, and nothing in the tree references the top-level `aioesphomeapi.fix_float_single_double_conversion`. I'll also flip `test_fix_float_helper_remains_public` to assert it is *not* exported (`not hasattr(aioesphomeapi, "fix_float_single_double_conversion")`) so the test pins the intended private state rather than the old leak.

aioesphomeapi/model.py
```diff
@@ -2066,3 +2066,164 @@ def build_unique_id(
...
+    "WaterHeaterState",
+    "WaterHeaterStateFlag",
+    "ZWaveProxyFeature",
+    "ZWaveProxyFrame",
+    "ZWaveProxyRequest",
+    "ZWaveProxyRequestType",
+    "build_unique_id",
+    "cached_fields",
+    "converter_field",
+    "message_types_to_names",
```
bdraco: cached_fields, converter_field, message_types_to_names, COMPONENT_TYPE_TO_INFO not intended to be public either

bdraco (changes requested):
cached_fields, converter_field, message_types_to_names, COMPONENT_TYPE_TO_INFO not intended to be public either

bdraco:
@esphbot rr

aioesphomeapi/model.py
esphbot: Agreed — all four are internal and should come out of `__all__`. Worth flagging the mechanics so this lands cleanly: dropping them from `__all__` only changes what `from .model import *` re-exports (i.e. the `aioesphomeapi.*` root). The real consumers use explicit imports that `__all__` doesn't gate — `from .model import message_types_to_names` in connection.py:57 and client.py:187, and `from aioesphomeapi.model import COMPONENT_TYPE_TO_INFO` in tests/test_model.py:87 — so no renames or call-site edits are needed; they keep working.

One dependency to handle in the same change: `test_model_all_matches_defined_names` asserts `set(model.__all__) == _model_public_defined_names()`, and since all four are public top-level definitions in model.py, removing them from `__all__` makes that equality fail. Reworking it to subtract an explicit `_NOT_EXPORTED = {"COMPONENT_TYPE_TO_INFO", "cached_fields", "converter_field", "message_types_to_names"}` keeps the "new public class silently not re-exported" guard while allowing these intentional exclusions.

aioesphomeapi/__init__.py
esphbot: Confirmed resolved in the current branch: `__init__.py` is identical to main (no `from .util import fix_float_single_double_conversion` line was added), the name is absent from `model.__all__`, and `test_fix_float_helper_not_public` now asserts `not hasattr(aioesphomeapi, "fix_float_single_double_conversion")`. model.py keeps its own internal import for the float converters, and tests/test_util.py still reaches it via `util.fix_float_single_double_conversion`, so nothing breaks — the only effect is that it's no longer reachable at the package root.

bdraco (approved):

PR #1791, opened 2026-06-23
Add DisconnectReason to DisconnectRequest

kbx81:

Mirrors the firmware `api.proto` change that adds a `reason` enum field to `DisconnectRequest` (`DISCONNECT_REASON_UNSPECIFIED` / `DISCONNECT_REASON_PROVISIONING_CLOSED`) and regenerates the protobuf bindings.

When a device requests the disconnect with a reason — e.g. its EN18031 provisioning window has closed — the connection parses it onto a new `disconnect_reason` attribute and logs it. The default is `UNSPECIFIED`, so older devices that omit the field are unaffected (backward-compatible proto3 addition). Surfacing the reason further up to Home Assistant's UI is left as a follow-up.

## Types of changes

- [ ] Other

- N/A

- esphome/esphome#17152

## Checklist:


esphbot (changes requested):
Blocking issues found.

- DisconnectReason.Name() raises ValueError on a reason the client doesn't know — aborts disconnect handling

kbx81:
Good catch — fixed in c83f512. `DisconnectReason.Name()` now only runs for known values (`msg.reason in DisconnectReason.values()`); an unknown reason from a newer device logs the raw integer instead of raising, so the `DisconnectResponse` send and `_cleanup()` always run. The raw value is still stored on `disconnect_reason`. Added `test_disconnect_request_with_unknown_reason` covering the unknown-value path (verified against a fresh Cython build).

esphbot (approved):
No blocking issues found.

aioesphomeapi/api.proto
```diff
@@ -158,6 +158,16 @@ message AuthenticationResponse {
   bool invalid_password = 1;
 }
 
+// Reason a party is requesting the connection be closed.
+enum DisconnectReason {
+  // No specific reason / not provided (default for older peers).
+  DISCONNECT_REASON_UNSPECIFIED = 0;
```
bdraco, line 164: should we add DEEP_SLEEP?

kbx81: Yes, Jesse and I had a chat about it and, in a follow-up PRs, we'll add more reasons. Should let us better tune how HA responds to disconnects. 😄

bdraco (approved):

PR #1800, opened 2026-06-25
Defer noise and cryptography imports until an encrypted connection

bdraco:

The noise frame helper pulls in `cryptography`, the `noise` protocol stack, and `chacha20poly1305_reuseable`, which are only needed for encrypted connections; they were being imported eagerly when the package was imported, and even a plaintext connection dragged them in because `packets.py` imported `make_noise_packets` from `noise_encryption`.

This moves `make_noise_packets` into `noise.py` so the plaintext packet builder no longer depends on the noise stack, and loads the noise frame helper lazily on the first encrypted connection. The cold import is run in an executor so it never blocks the event loop; once the module is cached it binds inline.

For a use case like `esphome logs` against a plaintext device, the whole noise and cryptography stack is now skipped; warm import of `aioesphomeapi` drops by about 12 ms locally and the stack is no longer resident. For encrypted devices the cost simply moves to connect time, which is already waiting on the network.

## Types of changes

- [ ] Other

- fixes <link to issue>

## Checklist:

### DIFF
--- a/aioesphomeapi/_frame_helper/noise.pxd
+++ b/aioesphomeapi/_frame_helper/noise.pxd
@@ -9,9 +9,20 @@ from .base cimport (
     _MAX_NAME_LEN,
 )
 from .noise_encryption cimport EncryptCipher, DecryptCipher
-from .packets cimport make_noise_packets
 
 cdef bint TYPE_CHECKING
+
+
+@cython.locals(
+    type_="unsigned int",
+    data=bytes,
+    data_header=bytes,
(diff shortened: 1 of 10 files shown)


esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

esphbot (approved):
No blocking issues found.

PR #1845, opened 2026-07-31
Add DeviceCapabilities message support

jesserockz:

Adds client support for the new `DeviceCapabilitiesRequest` / `DeviceCapabilitiesResponse` native-API messages (ids 149/150).

`DeviceInfoResponse` is served before authentication, so it should stay limited to identity information rather than growing an ever-longer flat list of optional-feature flags. `DeviceCapabilitiesResponse` is the new home for those flags and is only served on an authenticated (and, where configured, encrypted) connection. Each feature gets its own sub-message so it can gain fields over time without crowding the top-level field numbering: `BluetoothProxyCapabilities`, `VoiceAssistantCapabilities`, `ZWaveProxyCapabilities`, plus the existing `SerialProxyInfo` reused for `serial_proxies`.

This is not a breaking change. Devices still send all the old `DeviceInfoResponse` fields with the same values and will keep doing so until a future removal release, so older clients are unaffected.

Changes:

- `api.proto`: the `device_capabilities` rpc and the new message block, copied verbatim from esphome's `api.proto` so the two copies stay textually aligned. Regenerated `api_pb2.py` with the pinned `ghcr.io/esphome/aioesphomeapi-proto-builder` image. No existing `DeviceInfoResponse` field is marked deprecated, since devices are still sending them.
- `core.py`: registers 149/150 in `MESSAGE_TYPE_TO_PROTO`.
- `connection.py`: bumps the client's advertised API version in `HelloRequest` from 1.14 to 1.15, since the library now understands the 1.15 message set.
- `model.py`: `DeviceCapabilities` plus the three capability sub-models, following the existing `APIModelBase` / `converter_field` patterns. Feature-flag fields stay plain `int`, matching `DeviceInfo.bluetooth_proxy_feature_flags`; the `BluetoothProxyFeature` / `VoiceAssistantFeature` / `ZWaveProxyFeature` enums callers test against are unchanged.
- `client.py`: `APIClient.device_capabilities()` for the raw RPC, and `device_capabilities_compat(device_info)` which makes the RPC when the *device* reports API >= 1.15 and otherwise synthesises the same values from an already-fetched `DeviceInfo`. That fallback lives here next to the existing `*_feature_flags_compat(api_version)` helpers so consumers (Home Assistant first) do not each reimplement the version check and field mapping.

## Types of changes

- [ ] Other

- fixes <link to issue>

- esphome/esphome#17984

## Checklist:


esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - api.proto change has no linked esphome PR (repo hard rule)
> - Unknown api_version silently zeroes modern feature flags

esphbot (approved):
> [!TIP]

bdraco (approved):

Issue #1613, opened 2026-04-24
zeroconf exception hidden & fails to connect


stromnet:
I've been using aioesphomeapi for a [project](https://github.com/stromnet/esphome-prometheus-exporter), doing initial development on Linux. It uses the ReconnectLogic to directly connect to  node (no zeroconf used). It connected just fine to nodes and all working.

Then I tried it on FreeBSD, where I got some errors and no connections (ignore the actual error, it's not really what this ticket is about):

```
2026-04-24 06:30:16,396 DEBUG Trying to connect to garage-sensor @ 192.168.1.100
2026-04-24 06:30:16,397 DEBUG Connected lock acquired for garage-sensor @ 192.168.1.100
2026-04-24 06:30:16,397 DEBUG Starting zeroconf listener for garage-sensor
2026-04-24 06:30:16,397 DEBUG Creating new AsyncZeroconf instance
2026-04-24 06:30:16,397 ERROR Address in use when binding to ('', 5353); On BSD based systems sharing the same port with another stack may require processes to run with the same UID; When using avahi, make sure disallow-other-stacks is set to no in avahi-daemon.conf
```

Then nothing. The device is never connected to.
After some debugging it turns out to be the call to [_start_zc_listen()](https://github.com/esphome/aioesphomeapi/blob/c6ef4d677fc4271376bfbb3a75797be8caeb0264/aioesphomeapi/reconnect_logic.py#L345C1-L345C36) that throws an exception. This is however not logged as the connect_task is not monitored for exceptions.
With a simple task monitor like this one:

``` 
def monitor_task(task: Task[Any], logger):
    def _log_task_result(completed_task: asyncio.Task[Any]) -> None:
        try:
            exc = completed_task.exception()
        except asyncio.CancelledError:
            return

        if exc is not None:
            logger.exception("Background task %s failed", completed_task.get_name(), exc_info=exc)

    task.add_done_callback(_log_task_result)
    return task
```

I could capture the exception:
```
2026-04-24 06:32:17,343 ERROR Background task garage-sensor @ 192.168.1.100: aioesphomeapi connect failed
Traceback (most recent call last):
  File "/usr/home/johan/dev/esphome-prometheus-exporter/.pi-venv/lib/python3.11/site-packages/aioesphomeapi/reconnect_logic.py", line 381, in _connect_once_or_reschedule
    self._start_zc_listen()
[...]

PR #1870, opened 2026-08-16
Defer zeroconf loading until the mDNS wake listener is needed

bdraco:

Importing aioesphomeapi always pulled in the zeroconf package, which is the biggest piece of the import cost, about 90ms cold. When the dashboard hands `esphome logs` a pre-resolved IP via `--mdns-address-cache`, the happy path imported zeroconf, registered the mDNS listener, then tore it down as soon as the connection succeeded without it ever doing anything.

zeroconf now loads lazily. `ReconnectLogic` arms a 2s timer when a connect cycle starts; if the connection lands inside that window the timer is cancelled and zeroconf is never imported. If the attempt is still in flight at 2s the listener comes up mid-attempt, so a device that boots right after our SYN got dropped is woken by its boot announcement instead of waiting out a TCP timeout; a failed attempt starts the listener immediately for the backoff wait. The cold import runs in the executor so the event loop never blocks, same pattern as the noise stack loader. Since zeroconf requires a `RecordUpdateListener` subclass at registration, the listener is now a small delegate in `_zc_listener.py`; `ReconnectLogic` keeps its public surface including `async_update_records`.

Measured with `esphome logs` against a live device with a cache hit: time to the first connect attempt drops from 194ms to 170ms median, and the zeroconf stack stays completely unloaded on the happy path. Verified against an unreachable address that the listener still starts 2s into the attempt. `test_lazy_imports.py` now pins zeroconf and ifaddr as deferred.

Note: `aioesphomeapi.zeroconf.ZeroconfInstanceType` is now a type checking only alias and can no longer be imported at runtime; it was never re-exported from the package and no known consumer imports it. `ReconnectLogic.async_update_records` is removed as well; it existed only as the zeroconf listener callback, which now lives on the internal wake handler.

## Types of changes

- [x] Other

- fixes <link to issue>

## Checklist:

### DIFF
--- a/aioesphomeapi/_zc_listener.py
+++ b/aioesphomeapi/_zc_listener.py
@@ -7,21 +7,21 @@ from zeroconf import RecordUpdateListener
 if TYPE_CHECKING:
     from zeroconf import RecordUpdate, Zeroconf
 
-    from .reconnect_logic import ReconnectLogic
+    from .reconnect_logic import ZeroconfWake
 
 
 class ReconnectRecordUpdateListener(RecordUpdateListener):
-    """Forward zeroconf record updates to a ReconnectLogic.
+    """Forward zeroconf record updates to a ZeroconfWake.
 
     Lives in its own module so importing reconnect_logic does not import
     the zeroconf package; this module loads only when the mDNS-wake
     listener actually starts.
     """
 
-    def __init__(self, logic: ReconnectLogic) -> None:
-        self._logic = logic
+    def __init__(self, wake: ZeroconfWake) -> None:
+        self._wake = wake
 
     def async_update_records(
         self, zc: Zeroconf, now: float, records: list[RecordUpdate]
     ) -> None:
-        self._logic.async_update_records(zc, now, records)
+        self._wake.async_update_records(zc, now, records)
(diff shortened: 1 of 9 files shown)


esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - Deferring `_ptr_alias`/`_a_name` makes 10 tests order-dependent (CI runs `pytest -n auto`)

bdraco:
> question=None` means no cache replay, so the 2s deferral is a true blind window

Intended.  Cache is not prove of liveliness

esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - One test is still order-dependent: fails when run on its own

esphbot (approved):
> [!TIP]

bdraco:
> 1. Cold import has no per-loop lock, unlike the noise loader it is modelled on

HA will already have it loaded, esphome uses one per process. YAGNI

esphbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - Listener-failure warning is now DEBUG on the path that matters (issue #1613 signal lost)

esphbot (approved):
> [!TIP]

PR #1881, opened 2026-08-23
Build with Cython 3.3.0 by reconciling annotations with the pxd declarations

bdraco:

Cython 3.3.0 (released yesterday, and unpinned by `Cython>=3.2.9`) now analyses the annotations in a `.py` file against its `.pxd`, so the `use_cython` jobs on every open PR fail at build time. Three things in our code trip it.

`write_packets`, `make_plain_text_packets` and `make_noise_packets` annotate `packets: list[tuple[int, bytes]]` while the pxd declares `list packets`; Cython compares subscripted types structurally and a pxd has no spelling for the Python `int` inside the tuple, so the signatures never match. The annotation now goes through a `Packet = tuple[int, bytes]` alias that Cython cannot resolve, which makes it read the argument as a plain `list` again; mypy still sees the full type. The `type_: int` and `data: bytes` locals in those loops lose their annotations since the pxd `@cython.locals` already owns them and 3.3.0 reports the pair as a redeclaration.

**Breaking:** `ConnectionParams` is no longer a dataclass. Cython 3.3.0 rejects an annotated field with a default (`timezone: str | None = None`) on a cdef class declared in a pxd, and the one spelling it accepts silently drops the default. It stays a cdef class with the same attributes, constructor signature and defaults, but `dataclasses.replace`, `fields`, `astuple` and the generated `__eq__`/`__repr__` no longer work on it. The tests used `replace`; each fixture now builds its own params and sets the attribute directly. Keeping the dataclass and dropping the two defaults instead would also build, but `timezone=None` and `provide_time=True` are part of the public constructor, so that was rejected.

Verified by generating the Cython annotated HTML and C for 3.2.9 on main and 3.3.0 on this branch. On the same Cython version every `APIConnection`, `APIClientBase` and frame helper function is byte for byte identical, including the packet builders and `write_packets`; the only functions that differ are the `ConnectionParams` ones that changed on purpose. The per line annotation scores show no source line getting worse from this change; the remaining cross version deltas are Cython 3.3.0's own codegen. Full test suite passes on both the compiled and `SKIP_CYTHON` paths, mypy is clean.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

- fixes <link to issue>

## Checklist:

### DIFF
--- a/aioesphomeapi/_frame_helper/base.py
+++ b/aioesphomeapi/_frame_helper/base.py
@@ -12,6 +12,7 @@ if TYPE_CHECKING:
     from collections.abc import Callable, Iterable
 
     from ..connection import APIConnection
+    from .packets import Packet
 
 _LOGGER = logging.getLogger(__name__)
 
@@ -144,9 +145,7 @@ class APIFrameHelper:
         return cstr[original_pos:new_pos]
 
     @abstractmethod
-    def write_packets(
-        self, packets: list[tuple[int, bytes]], debug_enabled: bool
-    ) -> None:
+    def write_packets(self, packets: list[Packet], debug_enabled: bool) -> None:
         """Write a packets to the socket.
 
         Packets are in the format of tuple[protobuf_type, protobuf_data]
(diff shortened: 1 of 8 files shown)


esphbot (approved):
> [!TIP]

PR #1886, opened 2026-08-24
Restore null float fields as NaN in from_dict

bdraco:

JSON has no way to represent NaN or inf, so serializers like orjson write null for those floats. When a model that was serialized with `to_dict` is loaded back with `from_dict`, a float field that was NaN comes back as None instead of a float, which breaks callers that expect a float there. This maps null back to NaN for float fields in `from_dict` so a state cached to JSON restores with the same type it was saved with; Home Assistant needs this to persist deep sleep device states across restarts.

## Types of changes

- [ ] Other

- fixes <link to issue>

## Checklist:


esphbot (approved):
> [!TIP]

